PR #27, opened 2018-11-02
Add fabric version to doc

marcin-wosinek:


Manoj-nathwani:
Not required anymore after rewrite

PR #31, opened 2018-12-17
30 search tables

sebgrebe:
* Added search to profiles and groups list tables (using https://datatables.net/)
* There is only one search field for all columns, instead of individual search fields for each column as on old page (seemed overkill)
* Order of profiles is now by name rather than country
* Potential problem @Manoj-nathwani : 
   * Since there are quite a few users/groups in the database without any information beyond name, the top rows on both tables are entries with no location/link to community page. 
   * Should we filter out profiles (maybe also groups) without information for each column?


Manoj-nathwani:
Looks great! Lets jazz it up with the bootstrap theme:
- https://datatables.net/examples/styling/bootstrap.html

Otherwise it looks all a-ok 👌

PR #36, opened 2018-12-19
33 restyle map

sebgrebe:
* Add toggle between maps
* /profiles and /groups return now both profiles and groups data (see views.py) so that both maps can be displayed on both pages 
* Add [snazzy map "Flat pale"](https://snazzymaps.com/style/14889/flat-pale). Not particularly "snazzy", probably want to change to different one later
* Refactor some js and css into  individual files and folders

Concern:
* When toggle, map changes, but not list beneath it. Potentially confusing?


Manoj-nathwani:
Nice stuff!

Map on the homepage doesn't load

Some of the css I was overwriting on individual pages isn't being respected. You'll need to migrate that over to our own css file. For example the page width should be a lot wider and the logo smaller. Maybe you've just missed the css being overwritten on the `base.html` file actually 🤔 I guess you can simply compare the site locally to the production site to see differences to double check.

Manoj-nathwani:
oh! Can you also set it so `/profiles` defaults to the map on the `individual` mode and then `/groups` defaulting to the `groups` mode. I can help out with the backend logic for that if you want, though i'm surprised you managed to make good progress on editing the `views.py` files without any assistance already, so let's see what happens 😅  👏👏👏

PR #125, opened 2019-01-19
114 frontend fixes

sebgrebe:


sebgrebe:
fixes more issues:
- disappeared map
- wording in login/signup
- font color in navbar

PR #127, opened 2019-01-22
Create local development environment

taymonbeal:
Anyone can now clone the repo, run docker-compose up, and get a completely local and isolated environment with its own database.

IMPORTANT PRODUCTION NOTE: The environment variables listed under web in docker-compose.yml must be set in production before this is deployed.


Manoj-nathwani:
Nice one! Yeah this is deff needed, particularly the postgres database. I don't think the email stuff needs to act differently when testing though. Isn't it best to just hook into our prod SendGrid account to use that to test?

I'm getting the following error when running `fab run`:
`django.core.exceptions.ImproperlyConfigured: Requested setting LOGGING_CONFIG, but settings are not configured. You must either define the environment variable DJANGO_SETTINGS_MODULE or call settings.configure() before accessing settings.`

For context, i'm using the `fabfile.py` file so it may need changing to reflect the changes you've made.

Also getting profile pictures sorted it much higher priority, so you may want to pause this for now and sort that out.

taymonbeal:
The idea behind the email is that you shouldn't need _any_ production credentials just to clone the repository and run the app locally. Also, at least I personally find mailcatcher really convenient; you just open localhost:8001 and any emails you send appear there right away.

Did you try using `--build` to rebuild the image? You might be running the new code against an old image. In general it's necessary to do this whenever changes are made to the Dockerfile.

I don't see anything in the fabfile that looks like it should need to change. The thing that needs to change is the production environment variables, and those seem to live elsewhere.

This was a blocker for any other work on my end because my IP address wasn't whitelisted for the production DB. I've already forked this branch and the profile images work is based on it.

taymonbeal:
Actually I'm realizing there were a couple extra changes in my latest branch that were accidentally omitted from this one and are necessary for it to work. Force pull and try again?

Manoj-nathwani:
Ok sounds good!
Never used mailcatcher but it sounds good.
This means people from outside rthink can help contribute 🏆 

I managed to get it running, but i'm getting this error now:
`django.db.utils.ProgrammingError: relation "groups_group" does not exist
web_1   | LINE 1: ..._group"."lon", "groups_group"."edit_history" FROM "groups_gr...`

Presume this is because the local postgres db hasn't been set up?
I'd usually run a `python manage.py makemigrations` and then `migrate` however you've removed the file.

taymonbeal:
Just say `django-admin` instead of `manage.py`. The separate `manage.py` file was to set environment variables, but this is unnecessary with containers as it can be done in the Dockerfile.

Also, you shouldn't need to run `makemigrations`.

Ideally migrations would be run automatically on code changes but I haven't figured out a good way to make that work yet.

Manoj-nathwani:
IT WORKS!



I think the last time I worked in a team with a big docker django setup we just ran the migration command before the server started. That way it auto-migrates the prod db whenever the image uploads. It'll also fix the issue of having the run it manually when testing locally too.

- Can you edit the `fabfile.py` and replace in `django-admin` instead of using the `manage.py` file
- Other than that it's good to deploy 👍 
- Feel free to merge master and then ping me, i'll update the azure envs and deploy for you 😃

taymonbeal:
There already weren't any references to manage.py in the fabfile. There were some in the readme though, I'm editing those now after confirming that everything works.

By the way, where exactly are Azure environment variables controlled?

Issue #61, opened 2018-12-30
Default Hub Avatars


Richenda:
Here's a link for the stock images we should use for avatars (aka profile pictures) https://drive.google.com/drive/folders/1tUDZzlZF5whbcAvGgc2Mm8r3BVmK4gjf?usp=sharing
There are two categories: 'white' and 'colour'. If the space behind the picture will be white then we will use the coloured set. If the space behind is coloured we will use avatars that have the white fill.

I would like it if all Hub users are assigned one of these avatars randomly. They should then be given a choice in their user settings either to upload their own image OR to choose another stock avatar. In that section we will have all of these images (either all colour or all white) there for the users to pick from.

Richenda:
One more thing... is it safe posting a publicly viewable google drive link here? I need to protect the art.

sebgrebe:
@Richenda It's not if you don't want people to have access to it as the repo is public. However, once the site is up, anyone could download the pics anyway.

michaltrzesimiech:
@Richenda can we close this now?

PR #136, opened 2019-01-24
Add uploadable user profile images

### LINKED ISSUE #117: figure out how profile & group images will work

taymonbeal:
It's not pretty (in particular we need to figure out how to restore the attractive form styling), but it works.

IMPORTANT PRODUCTION NOTE: This adds production dependencies on Azure Blob Storage and Azure Cache for Redis, along with environment variables to configure them.

Fixes #117.


taymonbeal:
Yeah, we need to fix the HTML but we need to do it in a way that makes Django forms still work. I'll try applying `django-crispy-forms` but we might still need some frontend support on this. I'll let you know what the results wind up looking like.

Manoj-nathwani:
Updated the above feedback to move bugs i was getting for not following your instructions 😅

1) Totally agree, i've never used it before, but i've seen `django-crispy-forms` recommended everywhere so we should deff move over to it.
2) For now can you update your code to conform to the current way of building forms so we can pass this PR through. We can then make a new PR & feature update for upgrading the way to do form templating. Also means we can go live straight away with this update. This is awesome 🏆

taymonbeal:
I can try, it's not 100% straightforward because it's involving a new kind of form widget that we haven't used before.

taymonbeal:
Applied changes, they seem to work.

Manoj-nathwani:
@taymonbeal Happy for you to charge forward to merge & deploy this if you're happy it works fine locally for ya. What's happening with default profile too? This solves uploading photos, so I presume we should make another card for setting default images?

taymonbeal:
We already have #61 open for that.

Manoj-nathwani:
# 🏆 This is great 🏆

Issue #327, opened 2019-02-22
Enforce standards for commit messages


taymonbeal:

taymonbeal:
If we're using the squash workflow, we can't automatically enforce anything, but the need to do so is lesser because the maintainers directly control all commit messages on master. So we just need to be all on the same page about best practices there.

taymonbeal:
Fixed in #340.

taymonbeal:
On second thought, we should probably have a real workflow around this after all, with automated tooling of some kind.

PR #340, opened 2019-02-25
Specify commit message practices

### LINKED ISSUE #324: Create standards for how to merge PRs (squash vs. rebase)

taymonbeal:
This only works if all committers are on board with it, so I've added everyone in @rtcharity/tech as a reviewer. Please review at your earliest convenience.

Fixes #324 and #327.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,3 +56,30 @@ $ docker login eahub.azurecr.io
 >>> username: eahub
 >>> password: <see lastpass>
 ```
+
+# Commit Message Practices
+
+These are primarily for the benefit of maintainers, but all contributors are
+urged to follow them in order to make maintainers' lives easier.
+
+- In general, follow the practices outlined in
+  ["How to Write a Git Commit Message"](https://chris.beams.io/posts/git-commit/)
+  by Chris Beams.
+- As an exception to the above, do not manually wrap the body of a commit
+  message. The main reason for this is because our workflow depends on the
+  GitHub web interface, which doesn't provide an easy way to do this. It does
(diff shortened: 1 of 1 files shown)


taymonbeal:
Shouldn't. I've been doing this stuff the whole time, including for small updates.

michaltrzesimiech:
Fine by me!

taymonbeal:
I'm honestly not sure how worried we are about IP issues. I had us institute the DCO because before that the practice was to make volunteers sign IP assignment agreements with Rethink Charity, which I personally am not allowed to do (without a bunch of paperwork) and which is generally considered unfriendly for an open source project. It was agreed that the DCO was an adequate substitute for this. @michaltrzesimiech, do you have any thoughts?

"Rebase" here means rebasing all the commits in a PR onto master one at a time, as opposed to squashing them all into a single commit. I think we should consistently do one or the other, and a single commit on master per PR seems like the more common practice, and generally nicer since usually nobody's generally going to care about all those intermediate commits. Do you disagree? (Note that contributors can and should still rebase their own branches to resolve merge conflicts.)

sebgrebe:
> Do you disagree?

No, not at all, was just genuinely curious :) Thanks for the explanation

michaltrzesimiech:
> It was agreed that the DCO was an adequate substitute for this. @michaltrzesimiech, do you have any thoughts?

I remember this was the agreed upon solution. I'm happy to consult this with our lawyers if there's anything that could work better for any of us here.

taymonbeal:
You could ask them if they think we actually need this, and if not we could get rid of it since it's a minor speedbump to contributing. But not a big deal either way.

Issue #229, opened 2019-02-12
Use Django forms for rendering


taymonbeal:
See, e.g., https://docs.djangoproject.com/en/2.1/topics/forms/

This will make it possible to change the workings of models without writing a bunch more HTML.

django-crispy-forms is probably the easiest way to integrate this with Bootstrap.

sebgrebe:
This has been partly addressed [in this PR](https://github.com/rtcharity/eahub.org/pull/346).
After this PR, I believe whenever we implemented forms we tried to use django forms wherever that was easily possible.
@taymonbeal Do you think it's still necessary to go through all forms and try to decrease the use of customised solutions?

taymonbeal:
Yes, I think we should figure out definitively what the most idiomatic and maintainable way to specify forms in Crispy is, and then do that everywhere. Like, I think it has some facilities to specify forms in the Django code instead of mucking around with templates.

taymonbeal:
@sebgrebe, do you need my help in continuing this?

victor-yunenko:
We're planning a soft redesign in the future, during which the forms would definitely be updated, hence I guess we can close this ticket for now? Please feel free to reopen if needed.

PR #346, opened 2019-02-25
229 django forms

sebgrebe:
addresses #229 

Aim was to replace html for forms by django-crispy forms. Ideally, each form would just be displayed by {{ form | crispy }} in the html.

For a couple of forms, I could not do this because rendering specific fields adequately through Django-crispy-forms proved difficult. In those cases I rendered those fields through html and all others fields in this form through {{ form.FIELDNAME | as_crispy_field }}. Those forms were:
- edit_profile.html. Problematic field: image upload (looks terrible with django forms and I couldn't figure out a non-hacky way to fix that)
- edit_profile_cause_areas.html: Problematic fields: cause areas and giving pledges taken. In django forms, they display the id of the cause area or giving pledge, not their name. Probably fixable, but couldn't figure it out.
- edit_profile_career: Problematic field: Expertise. Same problem as in previous file. 
- login.html: Alert appeared above email and password. Had to render it through html to fix its position

Independent of whether this can be simplified further, future additional forms can be simply implemented through django-forms with this PR.

### DIFF
--- a/eahub/config/build_settings.py
+++ b/eahub/config/build_settings.py
@@ -23,4 +23,5 @@ STATIC_ROOT = "/static/"
 STATIC_URL = "/static/"
 STATICFILES_STORAGE = "whitenoise.storage.CompressedManifestStaticFilesStorage"
 
+# django-crispy-forms
 CRISPY_TEMPLATE_PACK = 'bootstrap4'
--- a/eahub/config/build_settings.py
+++ b/eahub/config/build_settings.py
@@ -12,6 +12,7 @@ INSTALLED_APPS = [
     "eahub.base.apps.BaseConfig",
     "eahub.localgroups.apps.LocalGroupsConfig",
     "eahub.profiles.apps.ProfilesConfig",
+    "crispy_forms",
 ]
 
 # Core settings: security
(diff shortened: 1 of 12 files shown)


eahub/profiles/forms.py
```diff
@@ -20,7 +20,7 @@ def save(self, commit=True):
 
     class Meta(CaseInsensitiveUsernameFieldCreationForm.Meta):
         model = User
-        fields = ['email']
+        fields = ['name', 'email']
```
taymonbeal: Why this change? Does the signup flow still work properly after making it? I ask because this is a `ModelForm` and the `User` model doesn't have a `Name` field, so I'm not sure whether it's right or wrong.

eahub/config/build_settings.py
```diff
@@ -21,3 +22,5 @@
 STATIC_ROOT = "/static/"
 STATIC_URL = "/static/"
 STATICFILES_STORAGE = "whitenoise.storage.CompressedManifestStaticFilesStorage"
+
+CRISPY_TEMPLATE_PACK = 'bootstrap4'
```
taymonbeal: Add a comment `# django-crispy-forms` above this line, to follow convention.

taymonbeal:
Note that #229 will need to remain open until the points that you've addressed here are solved (which I'll look into as soon as I get the chance). But this PR can be merged without that, once the above issues are taken care of.

eahub/profiles/forms.py
sebgrebe: This was necessary to make sure the name appears before the email on the form.

sebgrebe:
Can we merge this?

taymonbeal:
Yes, that's what it means for it to be approved.

sebgrebe:
> Yes, that's what it means for it to be approved.

Thought the reviewer merges

PR #357, opened 2019-03-02
249 users members groups

### LINKED ISSUE #249: Ability for users to become members of groups
# Problem
- Today we only have the ability for users to be `organisers` of groups which means they run them however there's no way of expressing that you are a `member` of EA London meaning you just attend events and are part of the community.

# Solution
- Give users the ability to become members of groups

sebgrebe:
* closes #249 
* aligns design of "available as speaker" selector with other selectors in community edit


eahub/profiles/forms.py
```diff
@@ -2,8 +2,14 @@
 from django import forms
 from django.db import models
 from .models import Profile, CauseArea, GivingPledge
+from ..localgroups.models import LocalGroup
 from ..base.models import User
 
+class CustomisedMultipleModelChoiceField(forms.ModelMultipleChoiceField):
+
+    def label_from_instance(self, obj):
```
taymonbeal: Why do we need to override this?

eahub/profiles/forms.py
```diff
@@ -61,18 +67,20 @@ class Meta:
             'open_to_job_offers',
         )
         labels = {
-            'open_to_job_offers': ('Open to job offers:')
+            'open_to_job_offers': ('Open to job offers')
         }
 
-
 class EditProfileCommunityForm(forms.ModelForm):
+    groups = CustomisedMultipleModelChoiceField(queryset=LocalGroup.objects.all(), required=False, label="EA Groups")
```
taymonbeal: When you look at the HTML of a form generated this way, does it include the database IDs of the groups? If so, we should prevent this by setting `to_field_name` to `slug`.

eahub/profiles/migrations/0004_auto_20190302_1907.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.1.7 on 2019-03-02 19:07
```
taymonbeal: Only one migration per PR, please, unless there's a specific reason it needs to be broken up.

Also, I think we should specify a [through model](https://docs.djangoproject.com/en/2.1/topics/db/models/#intermediary-manytomany) here. We don't need any extra fields now, but we're likely to later and it's hard to change.

eahub/profiles/models.py
```diff
@@ -166,6 +167,7 @@ class Profile(models.Model):
         enum.EnumField(GivingPledge), blank=True, default=list
     )
     subscribed_to_email_updates = models.BooleanField(default=False)
+    groups = models.ManyToManyField(LocalGroup, blank=True)
```
taymonbeal: Prefer `local_groups`. We're using this terminology throughout the codebase because Django already has a thing called "groups" that we're likely to wind up using later. Same throughout the rest of this file.

eahub/profiles/forms.py
sebgrebe: The groups were displayed as Objects (<Localgroup ...>), this way I could access their names instead. Followed [this example](https://stackoverflow.com/questions/3695754/django-customizing-display-of-modelmultiplechoicefield). 
I've now changed the name to 'CustomisedModelMultipleChoiceField', noticed that I got 'Model' and 'Multiple' the wrong way around

eahub/profiles/forms.py
sebgrebe: No, the html only shows the group's name
screen shot 2019-03-04 at 20 11 39

eahub/profiles/forms.py
taymonbeal: We don't need this, just add a `__str__` method to the model.

eahub/profiles/views.py
```diff
@@ -137,9 +138,15 @@ def edit_profile_community(request):
         form = EditProfileCommunityForm(request.POST, instance=request.user.profile)
         if form.is_valid():
             profile = form.save(commit=False)
+            profile.local_groups.clear()
             organisational_affiliations = request.POST.getlist('organisational_affiliations')
             profile.organisational_affiliations = organisational_affiliations
-            profile.save()
+            group_affiliations = request.POST.getlist('local_groups')
+            print(group_affiliations)
+            local_groups = LocalGroup.objects.filter(id__in=group_affiliations)
```
taymonbeal: I'm pretty sure this must be leaking database IDs. You should be able to solve this with [`to_field_name`](https://docs.djangoproject.com/en/2.1/ref/forms/fields/#django.forms.ModelMultipleChoiceField.to_field_name).

eahub/profiles/migrations/0002_profile_organisational_affiliations.py
```diff
@@ -1,21 +0,0 @@
-# Generated by Django 2.1.7 on 2019-02-28 20:05
```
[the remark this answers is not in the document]
sebgrebe: You asked me to have only one extra migration per PR, where i had 2. I deleted the 2 and run makemigrations again. Not sure what to do instead to satisfy your request, feel free to make the changes you think are necessary.

eahub/profiles/views.py
sebgrebe: I've just tried this:
```local_groups = forms.ModelMultipleChoiceField(queryset=LocalGroup.objects.all(), required=False, label="EA Groups", to_field_name="name")```
 but then it again shows with "LocalGroup" on the frontend.

The problem is that we're using django-crispy froms on the frontend, which doesn't give us direct control to the objects properties: The [example of to_field_name](https://docs.djangoproject.com/en/2.1/ref/forms/fields/#django.forms.ModelChoiceField.to_field_name) in the doc uses the field name, with django-crispy-forms all we do is putting in {{form.local_group}}, so we're not explicitly referencing the LocalGroup objects properties. I guess we could implement this without django-crispy forms, but what was then the point of adding it?

eahub/profiles/migrations/0002_profile_organisational_affiliations.py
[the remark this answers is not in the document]
taymonbeal: The problem is that this was an *already existing* migration, not one that you introduced.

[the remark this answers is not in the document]
sebgrebe: Right, sorry. Will try to roll that back

sebgrebe:
Branch has been rebased with new master. If you're happy with [this answer](https://github.com/rtcharity/eahub.org/pull/357#discussion_r262632889), this PR should be ready for merge, @taymonbeal

PR #358, opened 2019-03-03
Add inactivity status to group profiles and display this info on map …

### LINKED ISSUE #278: Make "inactive" groups visually distinct on the map

sebgrebe:
…and group page

closes #278 


eahub/base/views.py
```diff
@@ -42,10 +42,11 @@ def getGroupsData():
     rows = Group.objects.all()
     map_data = ''.join([
         '{' +
-            'lat: {lat}, lng: {lon}, label:"{name}", path: "{path}"'.format(
+            'lat: {lat}, lng: {lon}, label:"{name}", active:"{active}", path: "{path}"'.format(
```
taymonbeal: Shouldn't this be a boolean instead of a string?

eahub/localgroups/models.py
```diff
@@ -32,6 +32,7 @@ class LocalGroup(models.Model):
     facebook_page = models.URLField(blank=True)
     email = models.EmailField(blank=True)
     meetup_url = models.URLField(blank=True)
+    active = models.BooleanField(default=True)
```
taymonbeal: Would prefer for this to be named `is_active` so it's clearer that it's a boolean. This change will necessitate redoing the migrations.

Also, have you confirmed that it's possible to create a new inactive group?

sebgrebe:
> I think this PR also needs to include the changes to the Airtable migration. I can handle that part if you don't feel comfortable doing so.

Please do

eahub/base/views.py
sebgrebe: This is sent to Javascript, and Javascript can't read Python's Booleans (throws error "True is not defined" in console)

eahub/localgroups/models.py
sebgrebe: > Would prefer for this to be named is_active so it's clearer that it's a boolean. This change will > necessitate redoing the migrations.

Tbh, don't think it's worth the time. I think it should be clear to anyone who reads this.

> Also, have you confirmed that it's possible to create a new inactive group?

No. Why is this relevant?

sebgrebe:
Rebased with master. If you're happy with [my answer](https://github.com/rtcharity/eahub.org/pull/358#discussion_r) and could include the airtable migration, this should be ready for merge

PR #359, opened 2019-03-03
149 claim group

### LINKED ISSUE #149: Ability for users to claim group as their own
# Problem
- Currently if you are the organiser of `EA London` however aren't marked as it, there's no way of claiming it as my own.

# Solution
- On the group profile page add a `claim this group` button where users can request to be added as an organiser to the group in question.
- The backend then emails `admin@eahub.org` & any other organisers of that group to confirm adding the user as an organiser of the group.
- Once someone grants access an email is sent out to the user telling them `hey your request to be an organiser of EA London has been granted`

sebgrebe:
closes #149


eahub/localgroups/views.py
```diff
@@ -1,14 +1,37 @@
...
+def claim_group(request, slug):
+    group = get_object_or_404(Group, slug=slug)
+    if request.method == 'POST':
+        if request.user.is_anonymous:
+            return render(request, 'eahub/claim_group_error.html')
+        else:
+            subject = "EA Group claimed: {0}".format(group.name)
+            user_eahub_url = "https://{0}/profile/{1}".format(get_current_site(request).domain,request.user.profile.slug)
+            message = """ User <a href="{0}">{1}</a> wants to claim group {2}.
+                          Please check whether the user is eligible to claim this group
```
taymonbeal: Those leading whitespaces will appear in the email. One possible option is to move this text into a template and get it with [`render_to_string`](https://docs.djangoproject.com/en/2.1/topics/templates/#django.template.loader.render_to_string).

eahub/localgroups/views.py
```diff
@@ -1,14 +1,37 @@
...
+                          and let them know of the outcome by emailing them at {3}.
+                      """.format(user_eahub_url, request.user.profile.name, group.name, request.user.email)
+            send_mail(subject, message, request.user.email, ['admin@eahub.org'])
+            return render(request, 'eahub/claim_group_success.html')
+    else:
+        form = ClaimGroupForm()
+        return render(request, 'eahub/claim_group.html', {
```
taymonbeal: Similarly, I'm not convinced we need this as a separate page, as opposed to just a button on the group page.

sebgrebe:
> We don't want to actually merge this until organizership exists.
> 
> Also, should this be available for groups that have organizers, or should people just contact the organizers in those cases?

My feeling is that it shouldn't be and that the button should therefore only be visible for groups without organisers.

eahub/localgroups/views.py
```diff
@@ -1,14 +1,37 @@
...
 from .models import LocalGroup as Group
 
 def view_group(request, slug):
     group = get_object_or_404(Group, slug=slug)
+
     return render(request, 'eahub/group.html', {
         'group': group
     })
+
+def claim_group(request, slug):
```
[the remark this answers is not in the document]
sebgrebe: Both implemented

eahub/localgroups/views.py
```diff
@@ -1,14 +1,37 @@
...
+            return render(request, 'eahub/claim_group_error.html')
+        else:
+            subject = "EA Group claimed: {0}".format(group.name)
+            user_eahub_url = "https://{0}/profile/{1}".format(get_current_site(request).domain,request.user.profile.slug)
+            message = """ User <a href="{0}">{1}</a> wants to claim group {2}.
+                          Please check whether the user is eligible to claim this group
+                          and let them know of the outcome by emailing them at {3}.
+                      """.format(user_eahub_url, request.user.profile.name, group.name, request.user.email)
+            send_mail(subject, message, request.user.email, ['admin@eahub.org'])
+            return render(request, 'eahub/claim_group_success.html')
```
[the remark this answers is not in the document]
sebgrebe: removed page and added message instead

eahub/localgroups/views.py
sebgrebe: added template

eahub/localgroups/views.py
sebgrebe: added a button, which displays textfield with more info on same page

eahub/localgroups/views.py
```diff
@@ -1,14 +1,38 @@
...
 from .models import LocalGroup as Group
 
 def view_group(request, slug):
     group = get_object_or_404(Group, slug=slug)
+
     return render(request, 'eahub/group.html', {
         'group': group
     })
+
+@login_required(login_url=reverse_lazy('login'))
```
taymonbeal: Do you actually need the `login_url` parameter? Does it not default to the right one? Can we fix it if not?

eahub/base/static/scripts/main.js
```diff
@@ -41,3 +41,15 @@ var navbar = document.getElementById('navbar')
 menu_btn.addEventListener('click', function() {
   navbar.style.display = navbar.style.display == 'inline-block' ? 'none' : 'inline-block';
 })
+
+var claim_group = document.getElementById('claim_group')
```
taymonbeal: Which page(s) does this code run on? If it's just on the details page, then I'm confused about why we don't just control this server-side.

eahub/localgroups/views.py
```diff
@@ -1,14 +1,38 @@
...
+def claim_group(request, slug):
+    group = get_object_or_404(Group, slug=slug)
+    if request.method == 'POST':
```
taymonbeal: As-is, this will raise a 500 error on a GET request, because you don't actually return a response in that case. [`require_POST`](https://docs.djangoproject.com/en/2.1/topics/http/decorators/#django.views.decorators.http.require_POST) is the right way to solve this problem.

eahub/templates/eahub/claim_group.html
```diff
@@ -0,0 +1,15 @@
+{% extends 'form.html' %}
```
taymonbeal: Since this file is no longer used, you should delete it.

eahub/templates/eahub/claim_group_success.html
```diff
@@ -0,0 +1,11 @@
+{% extends 'base.html' %}
```
taymonbeal: Same.

eahub/templates/eahub/group.html
```diff
@@ -54,5 +54,44 @@ <h1>{{ group.name }}</h1>
       {% include 'eahub/maps/profile_page_map.html' with lat=group.lat|default_if_none:0 lon=group.lon|default_if_none:0 %}
     </div>
   </div>
+
+  <div class="claim_group_container">
+  {% if user.is_authenticated %}
+  {% if messages %}
```
taymonbeal: This should be in base.html, I think. We want messages to be supported everywhere.

eahub/localgroups/views.py
```diff
@@ -1,14 +1,38 @@
...
+        subject = "EA Group claimed: {0}".format(group.name)
+        user_eahub_url = "https://{0}/profile/{1}".format(get_current_site(request).domain,request.user.profile.slug)
```
taymonbeal: It'd be good to use [`get_absolute_url`](https://docs.djangoproject.com/en/2.1/ref/models/instances/#django.db.models.Model.get_absolute_url) here instead of manually stitching together a URL.

Also, it's technically possible for a user to not have a profile, and this code will raise a 500 error in that case; you need to handle it. The handling doesn't need to be especially pretty, though, because for now this should be rare.

eahub/localgroups/views.py
```diff
@@ -1,14 +1,38 @@
...
+        message = render_to_string('emails/claim_group.html', {
```
taymonbeal: You could just pass in `user` and `group` and do the lookups in the template.

eahub/base/static/scripts/main.js
```diff
@@ -41,3 +41,15 @@ var navbar = document.getElementById('navbar')
...
+var claim_group_toggle = document.getElementById('claim_group_toggle')
+var claim_group_togglers = document.getElementsByClassName('claim_group_toggler')
+
+for (let toggler of claim_group_togglers) {
+  toggler.addEventListener('click', function() {
+    console.log(claim_group.style.display)
```
taymonbeal: Don't leave this in the final code.

eahub/localgroups/views.py
```diff
@@ -1,14 +1,38 @@
...
+            'group_name': group.name,
+            'user_email': request.user.email
+        })
+        mail_managers(subject, message)
+        messages.success(
+            request,
+            ''' Thank you, we have received your request to claim this group.
+            Our admin team will send you an email once they have checked your request. ''',
+        )
+        return redirect('/group/{}'.format(group.slug))
```
taymonbeal: Same re: `get_absolute_url`

eahub/localgroups/views.py
```diff
@@ -1,14 +1,38 @@
...
+    if request.method == 'POST':
+        subject = "EA Group claimed: {0}".format(group.name)
+        user_eahub_url = "https://{0}/profile/{1}".format(get_current_site(request).domain,request.user.profile.slug)
+        message = render_to_string('emails/claim_group.html', {
+            'user_eahub_url': user_eahub_url,
+            'user_name': request.user.profile.name,
+            'group_name': group.name,
+            'user_email': request.user.email
+        })
+        mail_managers(subject, message)
```
taymonbeal: You can't pass HTML here. There are two options:

- Create two templates, one for HTML email and one for plain text, and use the `message` and `html_message` parameters.
- Create only the plain-text template and just use the `message` parameter.

The second option should suffice, I think, since this will only be sent to site staff.

Also, did you actually test this locally? Or did it not have HTML tags visible in the output when you did?

eahub/localgroups/views.py
```diff
@@ -1,14 +1,38 @@
...
+        messages.success(
+            request,
+            ''' Thank you, we have received your request to claim this group.
```
taymonbeal: Do newlines actually work in flash messages? I thought they didn't.

eahub/templates/eahub/group.html
```diff
@@ -54,5 +54,44 @@ <h1>{{ group.name }}</h1>
...
+  {% else %}
+  <div class="text-center">
+    <div class="btn-group btn-group-sm text-center" role="group" id="claim_group_toggle">
+      <button type="button" class="btn btn-default claim_group_toggler"><i class="fa fa-sitemap"></i> Claim this group</button>
+    </div>
+    <div id="claim_group" class="col-md-6 col-md-offset-3 alert alert-info hub-info" role="alert">
+      <button type="button" class="close claim_group_toggler">
+        <span aria-hidden="true">&times;</span>
+      </button>
+      <p>If you are the organiser of this group, you can claim it so that you can edit it.
```
taymonbeal: Do we need this verbiage? Seems like we could just have either a list of organizers, or else just one sentence "We don't know who organizes this group." followed by a button "I organize this group".

eahub/base/static/scripts/main.js
sebgrebe: Not quite sure what you mean. This runs on group.html and allows toggling the display of the info field about group claims (in interaction with the bootstrap functionality for closing that field)

eahub/localgroups/views.py
sebgrebe: you're right, works without

eahub/localgroups/views.py
sebgrebe: Do you mean get_current_site? How should I do it instead?

eahub/localgroups/views.py
sebgrebe: It was working locally without HTML tags. Was following [this example](https://stackoverflow.com/a/28476681). So I think it should be fine to leave this as it is?

eahub/localgroups/views.py
sebgrebe: It's working

eahub/templates/eahub/group.html
sebgrebe: made it shorter

eahub/localgroups/views.py
```diff
@@ -1,14 +1,39 @@
...
     return render(request, 'eahub/group.html', {
         'group': group
     })
+
+@login_required
+@require_POST
+def claim_group(request, slug):
+    group = get_object_or_404(Group, slug=slug)
+    subject = "EA Group claimed: {0}".format(group.name)
+    user_eahub_url = "https://{0}/profile/{1}".format(get_current_site(request).domain,request.user.profile.slug)
```
taymonbeal: This still 500s if the user has no profile.

eahub/localgroups/views.py
```diff
@@ -1,14 +1,39 @@
...
+    message = render_to_string('emails/claim_group.html', {
+        'user_eahub_url': user_eahub_url,
+        'user_name': request.user.profile.name,
+        'group_name': group.name,
+        'user_email': request.user.email
+    })
+    mail_managers(subject, message)
+    messages.success(
+        request,
+        ''' Thank you, we have received your request to claim this group.
```
taymonbeal: Confirmed that the newlines don't do anything, so don't include embedded newlines in this string.

eahub/templates/eahub/group.html
```diff
@@ -54,5 +54,42 @@ <h1>{{ group.name }}</h1>
       {% include 'eahub/maps/profile_page_map.html' with lat=group.lat|default_if_none:0 lon=group.lon|default_if_none:0 %}
     </div>
   </div>
+
+  <div class="claim_group_container">
+  {% if user.is_authenticated %}
```
taymonbeal: I changed my mind. Users should see the claim button even if they're not logged in. If somebody organizes a group we want them to be invited to claim it (which involves creating an account if they don't already have one).

eahub/localgroups/views.py
sebgrebe: Hm? I added require_post. When you sent a GET to this, you get "Method Not Allowed (GET)". This was what you asked for, wasn't it?

sebgrebe: Can't replicate this. When the user has no profile (i.e., isn't logged in), it sends me to the login

eahub/templates/eahub/group.html
sebgrebe: Made the button visible, but once you click it, it now says "You need to be logged in to claim a group". If they're not logged in, we don't have their email address, so there's no point in letting them claim it

sebgrebe:
Has been rebased with master and implemented requested changes or gave answers where not. The only one I couldn't really address is [this](https://github.com/rtcharity/eahub.org/pull/359#discussion_r262218728). Feel free to make your own changes for this one. Should be ready to merge after

eahub/localgroups/views.py
taymonbeal: When you click on "My profile" as admin you get a 404, not a 500.

PR #402, opened 2019-03-12
306 other fields

### LINKED ISSUE #306: Add the "Other" free text field under "Expertise", "Cause areas", "Topics" to "Available as a speaker"

sebgrebe:
closes #306


eahub/profiles/migrations/0004_auto_20190312_1918.py
```diff
@@ -0,0 +1,28 @@
+# Generated by Django 2.2b1 on 2019-03-12 19:18
```
taymonbeal: Please rename this file to something more informative. `0004_profile_other_fields.py` would be good if you don't happen to have any better ideas.

sebgrebe: done

sebgrebe:
> I'm not 100% clear on what the relationship between the Other checkbox and the text box is. Can you clarify how that works?

This was a mistake, I removed the checkbox. 
Alternatively, we could keep the other checkbox and let it toggle the textarea (this is how it was before), but removing the checkbox was quicker. Can still implement the other checkbox toggling the textarea at some point (maybe not high priority)

eahub/profiles/models.py
```diff
@@ -66,7 +67,7 @@ class ExpertiseArea(enum.Enum):
         PSYCHOLOGY: "Psychology",
         PHYSICS: "Physics",
         MEDICINE: "Medicine",
-        OTHER_SCIENCE: "Other Science",
+        OTHER_SCIENCE: "Other Science"
```
taymonbeal: This looks like an accidental change. Revert?

eahub/profiles/models.py
```diff
@@ -168,12 +180,15 @@ class Profile(models.Model):
     cause_areas = postgres_fields.ArrayField(
         enum.EnumField(CauseArea), blank=True, default=list
     )
+    cause_areas_other = models.TextField(null=True, blank=True)
```
taymonbeal: `CharField`s and `TextField`s should not set `null=True` except in special circumstances, which this isn't. In Django, empty fields of those types are generally represented in SQL as the empty string, not as `NULL`.

This change will require deleting your migration file and rerunning `makemigrations`.

eahub/profiles/models.py
```diff
@@ -153,6 +154,17 @@ class OrganisationalAffiliation(enum.Enum):
         WILD_ANIMAL_INITIATIVE: "Wild Animal Initiative"
     }
 
+def create_pretty_property_list(propertyClass, standard_list, other_list):
```
taymonbeal: Rename to `property_class`.

taymonbeal: Nit: This isn't really "creating" anything (in particular it's not creating a database record), it's just formatting some text. Maybe `pretty_property_list` or `prettify_property_list`?

eahub/profiles/models.py
```diff
@@ -153,6 +154,17 @@ class OrganisationalAffiliation(enum.Enum):
...
+    pretty_list = ''
+    if standard_list:
+        pretty_list += ", ".join(map(propertyClass.label, standard_list))
+    if other_list:
+        pretty_list = pretty_list + ', ' + other_list if standard_list else other_list
+    if (standard_list and other_list) is False:
```
taymonbeal: `if not (standard_list and other_list):` (`is` in Python compares object identity and I'm not sure that does the right thing here)

sebgrebe:
> So there's no "Other" checkbox? Probably you should open a separate issue for creating one that causes the text box to appear or disappear if it's checked or unchecked, since I think @michaltrzesimiech wanted this. Note that this should be _only_ a UI change, not a database schema change; the database schema should remain the way you've got it now.

[created](https://github.com/rtcharity/eahub.org/issues/405)

sebgrebe:
> Also, you've got DCO problems, please fix them.

done

eahub/profiles/models.py
sebgrebe: Didn't implement as it change the functions behaviour (returns N/A when it shouldn't)

taymonbeal: This is definitely suspicious and I remain confused about how it works, but we can leave it for now.

eahub/profiles/models.py
```diff
@@ -168,12 +180,15 @@ class Profile(models.Model):
...
     )
+    cause_areas_other = models.TextField(blank=True)
     available_to_volunteer = models.BooleanField(null=True, blank=True, default=None)
     open_to_job_offers = models.BooleanField(null=True, blank=True, default=None)
     expertise_areas = postgres_fields.ArrayField(
         enum.EnumField(ExpertiseArea), blank=True, default=list
     )
+    expertise_areas_other = models.TextField(blank=True)
     available_as_speaker = models.BooleanField(null=True, blank=True, default=None)
+    topics_i_speak_about = models.TextField(null=True, blank=True)
```
taymonbeal: This is still `null=True`.

sebgrebe: done

PR #412, opened 2019-03-14
Send group related emails to lean@eahub.org 

### LINKED ISSUE #411: Customise who gets group claim emails
These are currently going to admins, but @Richenda should also get them. Devs (@taymonbeal , @sebgrebe) actually don't need to get them.

sebgrebe:
closes #411 

### DIFF
--- a/eahub/config/settings.py
+++ b/eahub/config/settings.py
@@ -25,7 +25,6 @@ ADMINS = list(env.dict("ADMINS").items())
 DEFAULT_FROM_EMAIL = "EA Hub <noreply@eahub.org>"
 EMAIL_SUBJECT_PREFIX = "[EA Hub] "
 MANAGERS = ADMINS
-GROUP_ADMIN = 'lean@eahub.org'
 SERVER_EMAIL = DEFAULT_FROM_EMAIL
 
 # Core settings: file uploads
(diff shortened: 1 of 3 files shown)


eahub/config/settings.py
```diff
@@ -25,6 +25,7 @@
 DEFAULT_FROM_EMAIL = "EA Hub <noreply@eahub.org>"
 EMAIL_SUBJECT_PREFIX = "[EA Hub] "
 MANAGERS = ADMINS
+GROUP_ADMIN = 'lean@eahub.org'
```
taymonbeal: This setting should be in the Local Groups section, and it should be an environment variable with a corresponding `docker-compose.yml` entry (which should be something like `groupadmin@example.com`).

sebgrebe: done

sebgrebe:
Added a LEAN_MANAGERS variable to azure, so this should be ready to go

eahub/localgroups/views.py
```diff
@@ -75,7 +76,7 @@ def claim_group(request, slug):
         'group_url': "https://{0}/group/{1}".format(get_current_site(request).domain,group.slug),
         'user_email': request.user.email
     })
-    mail_managers(subject, message)
+    send_mail(subject, message, settings.DEFAULT_FROM_EMAIL, recipient_list=settings.LEAN_MANAGERS)
```
taymonbeal: You have to pass a list of strings in `recipient_list`, not a list of tuples.

I think the easiest solution is to change it to:
`LEAN_MANAGERS = env.list("LEAN_MANAGERS")`
`LEAN_MANAGERS: leanmanager@example.com`

with no associated name. Apologies for not realizing earlier that this would be a problem.

sebgrebe: Initially I put in just a string and it came back with "recipient_list needs to be either list or tuple". And this is working, I'm receiving the emails on 8001

PR #414, opened 2019-03-14
393 group hyperlink on profile

### LINKED ISSUE #393: Make EA Groups in 'Community' hyperlinks
screenshot 2019-03-07 at 13 31 09
So when you click it takes you to the group profile.

sebgrebe:
closes #393

### DIFF
--- a/eahub/templates/eahub/profile.html
+++ b/eahub/templates/eahub/profile.html
@@ -93,10 +93,15 @@
           {% endif %}
           {% if profile.local_groups.exists %}
           <div class="list turquoise-links">
-            <div class="tag">Local groups:</div>
-            {% for group in profile.local_groups.all %}
-              <a href="{% url 'group' group.slug %}">{{ group.name }}</a>
-            {% endfor%}
+            {% if profile.local_groups.all.count == 1 %}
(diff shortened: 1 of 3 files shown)


eahub/templates/eahub/profile.html
```diff
@@ -92,7 +92,17 @@ <h3>Community</h3>
           <div class="list"><div class="tag">Organisational affiliations:</div> {{ profile.get_pretty_organisational_affiliations }}</div>
           {% endif %}
           {% if profile.local_groups.exists %}
-          <div class="list"><div class="tag">Local groups:</div> {{ profile.get_pretty_local_groups }}</div>
+          <div class="list turquoise-links">
+            {% if profile.local_groups.all.count == 1 %}
```
taymonbeal: There's an easier way. https://docs.djangoproject.com/en/2.1/ref/templates/builtins/#pluralize

sebgrebe: The point of the if/else is not the pluralising of "Local group/s", but rather to have a block list where no. of groups > 1 and an inline statement where no. of group = 1

PR #430, opened 2019-03-16
423 profile marker cluster disappear

### LINKED ISSUE #423: Marker cluster disappear on profiles map
> How come I can't see individual markers when I zoom in on the map? Do individuals work differently from groups in this way?

sebgrebe:
closes #423

### DIFF
--- a/eahub/base/static/scripts/maps/heatmap.js
+++ b/eahub/base/static/scripts/maps/heatmap.js
@@ -42,13 +42,13 @@ function renderMap(selectedMap, mapDataProfiles, mapDataGroups) {
 
 function renderProfileMap(locations) {
   var map = createMap();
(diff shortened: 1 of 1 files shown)


eahub/base/static/scripts/maps/heatmap.js
```diff
@@ -71,8 +72,7 @@ function createMap() {
   return map
 }
 
-function addMarkersWithLabels(locations, map) {
-
+function addMarkersWithLabels(locations, map, checkGDPR = false) {
```
taymonbeal: Wouldn't it make more sense to just get rid of `checkGDPR` entirely?

sebgrebe: yes. Branch updated

Issue #419, opened 2019-03-16
Private profiles


taymonbeal:
Users should be able to uncheck a box to have their profile not appear.

PR #436, opened 2019-03-17
Allow profiles to be marked private

taymonbeal:
This also moves around some of the privacy verbiage (since not everything is public anymore) and reworks how the organiser form displays candidates (because we can't just show everybody).

Newly created profiles default to public (with a checkbox on the initial form). Profiles imported from the legacy Hub default to private.

People should manually test this thoroughly.

Fixes: #419


eahub/base/views.py
```diff
@@ -61,8 +61,8 @@ def getGroupsData():
         'map_data': map_data
     }
 
-def getProfilesData():
-    rows = Profile.objects.all()
+def getProfilesData(user):
+    rows = Profile.objects.visible_to_user(user)
     map_data = ''.join([
         '{' +
             'lat: {lat}, lng: {lon}, label:"{name}", path: "{path}", gdpr_confirmed: "{gdpr_confirmed}"'.format(
```
sebgrebe: as we're moving to "private", the gdpr_confirmed field in the map_data hash can be removed.

eahub/localgroups/forms.py
```diff
@@ -1,45 +1,94 @@
...
+        if local_group is None or local_group.pk is None:
+            already_selected = models.Value(False, output_field=models.BooleanField())
+        else:
+            already_selected = models.Case(
+                models.When(pk__in=local_group.organisers.all(), then=True),
+                default=False,
+                output_field=models.BooleanField(),
+            )
+        queryset = base_models.User.objects.select_related("profile").annotate(
+            already_selected=already_selected
```
sebgrebe: I assume if a user is annotated as already selected, this does not block them from being selected as organiser for another group? Don't fully comprehend the code, so just want to check that's the case. I assume we want users to be able to be organisers of different groups

taymonbeal: `already_selected` is an annotation that only exists within a particular query, not a persistent database field. This query is only concerned with the particular group that's being edited and is completely independent of whatever's going on in any other groups.

Issue #420, opened 2019-03-16
Clustered anonymous profiles on the map


taymonbeal:
The map needs to be able to list, for a given location, the number of anonymous people located there.

taymonbeal:
Here's the query to use:

```python
Profile.objects.filter(lat__isnull=False).values('lat', 'lon').annotate(count=Count('*')).filter(count__gte=15).order_by()
```

PR #437, opened 2019-03-17
420 cluster anonymous profiles

sebgrebe:
partly addresses #420 

Profiles in one location are now available in list on label, rather than spiderfied. Did not apply this to groups as groups are rarely in same position and if they are, we want markers to spiderfy so that user can visually distinguish between active and inactive groups

Once number of anonymous profiles at a location is made available to frontend, this can be easily added to the profiles_at_location variable in heatmap.js to add anonymous profiles to clusters of profiles, which would the fully fix 420.


taymonbeal:
It's probably simplest to hold off on merging this until after #436 is in.

PR #439, opened 2019-03-17
Add other group type

### LINKED ISSUE #390: Add "Other" to "Group Type"
image

We have at least 1 corporate group and discuss regional groups. This can get useful and asked for quickly.

sebgrebe:

closes #390


sebgrebe:
Done.
I've also changed the display of the organisers such that where they are unknown or there is only one, it's not displayed as a list - looks better

Issue #421, opened 2019-03-16
Figure out how to incorporate password reset links into launch email


taymonbeal:

taymonbeal:
Michal, let me know when you have a template that a password reset link could go into.

Manoj-nathwani:
@michaltrzesimiech, See https://github.com/rtcharity/eahub.org/pull/450#issuecomment-474202263

michaltrzesimiech:
The template itself is now in our MailChimp. I'll work on the text later today.

taymonbeal:
I don't have access to this, can you share?

michaltrzesimiech:
Sorry, I was meaning to do that before but got distracted. Sent you an invite to join as a user.

taymonbeal:
We're going to use a Django management command for this. Still need to figure out how to handle the email addresses that aren't in the legacy database.

michaltrzesimiech:
There will only be emails that are in either prod or the legacy database

taymonbeal:
Folding this into #514.

PR #450, opened 2019-03-18
autofill reset email field

Manoj-nathwani:
# Problem
- Issue https://github.com/rtcharity/eahub.org/issues/421
- We need a way to auto-fill the email field on the reset-passoword page so we can send mailshots to users to reset their password with little additional effort required on their end

# Solution
- Added some simple javascript to extract the email address from the url like `https://eahub.org/profile/password_reset/#hello@email.com` and fill in the input field

# Further steps
- Update the email we're sending to old users to include the `#hello@email.com` at the end of the URL

### DIFF
--- a/eahub/templates/registration/password_reset_form.html
+++ b/eahub/templates/registration/password_reset_form.html
@@ -13,3 +13,11 @@
 </form>
 </div>
 {% endblock %}
+
+{% block scripts %}
+  <script>
+    // Extract email address from https://eahub.org/profile/password_reset/#hello@email.com
+    // and set it to the email address input box (used for personalised reset password emails)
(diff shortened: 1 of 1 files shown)


Manoj-nathwani:
@taymonbeal Yeah, it seemed like any more complexity would require a lot more work. This seems like a dumb & simple solution to solving our mass email mailshot problem.

Manoj-nathwani:
@Richenda / @michaltrzesimiech - you're now free to update the mail shot we're sending to add a reset password button with the following url `https://eahub.org/profile/password_reset/#hello@email.com`

Notice how the end includes the email address of the user. You'll have to set that using whatever email platform you're using.

PR #494, opened 2019-03-25
change footer

### LINKED ISSUE #468: Specify terms for third-party reuse of data
Other people might wonder whether they can reuse the data that's public on the EA Hub for whatever purposes.

We need to figure this out pre-launch because whatever terms we have for this need to be binding on users who submit data.

sebgrebe:

closes #468, in particular [this](https://github.com/rtcharity/eahub.org/issues/468#issuecomment-476038917)

### DIFF
--- a/eahub/templates/base.html
+++ b/eahub/templates/base.html
@@ -118,7 +118,7 @@
     <footer class="footer">
       <div class="container">
         <strong>&copy; {% now "Y" %} Effective Altruism Hub.</strong>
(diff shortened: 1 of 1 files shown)


michaltrzesimiech:
I think we're ok with it, and we can either wrap that text (unless @Richenda objects) or decrease the font size, or both. Otherwise, it would be difficult to specify clearly that we only have the codebase licensed under MIT, but not the database. If we only mention "Licensed under MIT" and then specify this elsewhere, I think it isn't as clear from the legal standpoint (but I'm not a lawyer).

taymonbeal:
I'll merge this for now.

PR #506, opened 2019-03-26
Add number of profiles to marker

### LINKED ISSUE #491: Profile map shows counts of cities, not of individuals
This makes it look as though we have only a handful of people on the map, even though in fact we have hundreds.

sebgrebe:

adds number of profiles to marker on map, as suggested [here](https://github.com/rtcharity/eahub.org/issues/491#issuecomment-476442861)

closes #491

### DIFF
--- a/eahub/base/static/scripts/maps/heatmap.js
+++ b/eahub/base/static/scripts/maps/heatmap.js
@@ -134,15 +134,20 @@ function addMarkersWithLists(location_clusters, map) {
 
 function addDescription(marker, profiles) {
   public_profiles = profiles.filter(profile => !profile.anonymous)
+  private_profiles = profiles.filter(profile => profile.anonymous)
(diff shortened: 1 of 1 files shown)


sebgrebe:
Screen Shot 2019-03-26 at 16 33 55

taymonbeal:
Hmm. Do you think we could instead add an additional item to the end of the list that says "15 anonymous EAs" (or however many)? That strikes me as the least confusing way to handle the fact that we have two different kinds of things on the map.

sebgrebe:
> Hmm. Do you think we could instead add an additional item to the end of the list that says "15 anonymous EAs" (or however many)? That strikes me as the least confusing way to handle the fact that we have two different kinds of things on the map.

Good idea, updated:

Screen Shot 2019-03-26 at 19 58 49

PR #528, opened 2019-03-31
491a profile map

sebgrebe:
fixes [this error](https://github.com/rtcharity/eahub.org/issues/491#issuecomment-476974981)

### DIFF
--- a/eahub/base/static/scripts/maps/heatmap.js
+++ b/eahub/base/static/scripts/maps/heatmap.js
@@ -133,9 +133,11 @@ function addMarkersWithLists(location_clusters, map) {
 }
 
 function addDescription(marker, profiles) {
-  public_profiles = profiles.filter(profile => !profile.anonymous)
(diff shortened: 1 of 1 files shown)


eahub/base/static/scripts/maps/heatmap.js
```diff
@@ -146,13 +147,20 @@ function addDescription(marker, profiles) {
       marker.desc += '<li>' + private_profiles.length.toString() + ' anonymous ' + user_word + '</li>'
     }
     marker.desc += '</ul>'
-  } else if (public_profiles.length == 1) {
+  } else if (!exists(public_profiles) && exists(private_profiles)) {
```
taymonbeal: You don't need an `exists` function, this can just be `} else if (!public_profiles.length && private_profiles.length) {`.

sebgrebe: I just thought it reads easier

Issue #461, opened 2019-03-19
How should we treat links in custom text fields?


michaltrzesimiech:
Should we allow links (i.e. by validating against a regex)? If so, should we automatically turn them into hyperlinks?

![image](https://user-images.githubusercontent.com/14195560/54608067-3ba4b580-4a50-11e9-9393-75361a6be7ca.png)

taymonbeal:
This is part of a bigger question around text formatting.

Do we envision people ever entering multiple paragraphs here? I think right now the UI allows it but the page layout isn't very optimized for it. Maybe this is different for different fields?

taymonbeal:
Bumping the priority of this because it has implications for legacy profiles, some of which include rich text.

sebgrebe:
Allowing for link seems sensible (if this is not a security issue). Turning all links on the profile page into hyperlinks is two lines of code.

sebgrebe:
Made a PR that adds the hyperlinks in case we want to go with this: #470

taymonbeal:
The way im

taymonbeal:
The way I'm currently leaning towards handling this is allowing Markdown in multiline text fields. Do people feel that this makes sense?

sebgrebe:
I guess this would conflict with automatically turning URLs (using linkify) into links as linkify would turn the markdowned URLS into html, wouldn't it?

I think one disadvantage of using markdown is that most users won't be familiar with the syntax?

taymonbeal:
Markdown automatically turns URLs into links.

sebgrebe:
I know, but if we also have linkify, then that creates a mess (I think). What I'm saying is that we can't have both and if we implement markdown and not linkify and people don't know how to use markdown, then we don't have linkified URLs.

taymonbeal:
If they just enter URLs and they get Markdown processed then they'll appear as links.

sebgrebe:
I thought with markdown you need to do it like this:
```[name of site](www.site.com)```
or is that only github's markdown?

sebgrebe:
www.test.com

taymonbeal:
Yes, that's how you make a link with custom text, but if you just enter a raw URL Markdown renders it as a link too. In general Markdown is designed to do something reasonable even if users aren't familiar with the syntax.

Although it actually looks like not all Markdown implementations do this, so we'd want to use one that does.

sebgrebe:
[...]

PR #626, opened 2019-04-17
Render URLs in free-text fields as links

taymonbeal:
Fixes: #461


michaltrzesimiech:
Shouldn't we implement spam protection first before allowing this?

taymonbeal:
I don't know. We do use `rel=nofollow` so that mitigates things. Do we generally have the spam problem under control? We aren't going to have automated spam filtering until after the freeze but we could put this on ice until then if desired.

sebgrebe:
> I think this isn't necessary as linkify already implements this. Every div that has linkify as a CSS class turns URLs into links. So for example, the community panel on a profile page has linkified URLs because the div encompassing all panels has linkify as a CSS class.

@taymonbeal, what do you think about this? Can we close this PR?

taymonbeal:
I think it's preferable to do this on the server; the server-side implementation implements nofollow, which we need for spam reasons, and also doesn't slow down the page by forcing the user's browser (which might be a low-end smartphone) to do all the work. I've added a commit to remove linkify from the codebase.

taymonbeal:
Fixed.

PR #633, opened 2019-04-19
Order profiles and groups on list such that blanks last

### LINKED ISSUE #124: Sorting on lists puts blanks last
[Suggested](https://docs.google.com/spreadsheets/d/1qotdbleP9NHHieVsFkO2LQ6qeI43wBF2P0m_2NOe-xw/edit?ts=5c43311d#gid=0) by Marcin

sebgrebe:
closes #124

### DIFF
--- a/eahub/base/views.py
+++ b/eahub/base/views.py
@@ -83,7 +83,7 @@ def getGroupsData():
         if x.lat and x.lon
     ]
     return {
-        'rows': rows,
+        'rows': order_blank_last(rows),
         'map_data': map_data
     }
 
@@ -100,7 +100,7 @@ def getProfilesData(user):
         if x.lat and x.lon
     ]
     return {
-        'rows': rows,
+        'rows': order_blank_last(rows),
         'map_data': map_data
     }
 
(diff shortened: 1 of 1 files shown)


sebgrebe:
> Sorting is a lot faster if you do it in the database using an `order_by` clause.

I looked into this, but could not find a way to use order_by that allows for putting blanks last. 

If there shouldn't be an (easy to find) way to do this with order_by and you don't want to do the ordering server-side, we could just ditch this ticket all together. There are no public profiles or groups that are blank and one cannot add any new blank groups or profiles. Only worry would be old blank profiles that will be made public. Have these maybe been purged anyway when Michael checked for spam profiles?

taymonbeal:
Old spam profiles are unlikely to be made public. I think it's fine to ditch this ticket, or to postpone it until later. `order_by` is definitely possible but might be a little convoluted.

sebgrebe:
> Old spam profiles are unlikely to be made public. I think it's fine to ditch this ticket, or to postpone it until later. `order_by` is definitely possible but might be a little convoluted.

I'll look into the order_by solution, but let's move it to a later milestone (Plato)

Issue #634, opened 2019-04-19
On signin, "successfully signed in" shows up on group page


sebgrebe:
This message should be overridden

sebgrebe:
However, not getting this locally. I assume the PR that fixed this has not been deployed?

sebgrebe:
fixed after deployment

PR #574, opened 2019-04-08
558 auth pop ups

sebgrebe:
- fixes #558 
- simplifies overriding of allauth templates
- fixes #556

WARNING: This touches all authentication routes (login, signup, password reset; all urls can be found [here](https://github.com/pennersr/django-allauth/blob/master/allauth/account/urls.py)). I have tested all of them manually, but since we're now in post-launch time, it might be worth for whoever reviews this to do some selected tests. (A kingdom for some e2e on these user flows)

PR #635, opened 2019-04-19
fix url for group actions

sebgrebe:
I noticed that master branch and product are out of sync since [there is an error in production that doesn't show up on master](https://github.com/rtcharity/eahub.org/issues/634). This error was fixed in [this PR](https://github.com/rtcharity/eahub.org/pull/574). 

Furthermore, on master, I'm getting 404s when reporting groups as inactive or claiming them. However, this error doesn't show up in production. Again, this is because master and production are out of sync. This PR fixes the code that causes the error in master. Once this PR is merged, we should deploy.


taymonbeal:
Deploying now.

PR #644, opened 2019-04-22
352 report abuse

### LINKED ISSUE #352: Add "Report abuse" button to personal and group profiles

sebgrebe:
fixes #352 

- Allows logged in and not logged in users to report abuse
- Gives users the choice between three options: Spam, fake account, offensive content
- Send email to admin on report
- Reporters are anonymous, admins are not told who reported


eahub/base/static/scripts/report_buttons.js
```diff
@@ -10,6 +10,12 @@ let report_group_inactive = {
   togglers: document.getElementsByClassName('report_group_inactive_toggler')
 }
 
+let report_profile_abuse = {
```
taymonbeal: What does this code do?

eahub/localgroups/urls.py
```diff
@@ -18,4 +19,14 @@
...
         name="localgroups_delete",
     ),
+    urls.path(
+        '<slug:slug>/report-abuse/',
+        views.report_abuse,
+        name='report_abuse_group'
+    ),
+    urls.path(
+        '<slug:slug>/report-abuse-done/',
+        TemplateView.as_view(template_name='eahub/report_abuse_done.html'),
```
taymonbeal: Would it maybe be better for this to redirect back somewhere useful with a flash message instead of having a separate landing page?

eahub/localgroups/views.py
```diff
@@ -116,3 +118,15 @@ def report_group_inactive(request, slug):
         ''' Thank you, we have received your report. Our admin team will send you an email once they have looked into it. ''',
     )
     return redirect('/group/{}'.format(group.slug))
+
+
+def report_abuse(request, slug):
```
taymonbeal: This should be a [`FormView`](https://docs.djangoproject.com/en/2.2/ref/class-based-views/generic-editing/#formview).

eahub/profiles/views.py
```diff
@@ -173,3 +180,38 @@ def delete_profile(request):
         return render(request, 'eahub/delete_profile.html', {
             'form': form
         })
+
+class ReportAbuse:
```
taymonbeal: Per the above recipe, this logic should be part of the form class.

eahub/profiles/views.py
```diff
@@ -173,3 +180,38 @@ def delete_profile(request):
...
+    def __init__(self, reportee, type):
+        self.reportee = reportee
+        self.type = type
+
+    def send(self, request):
+        reasons = request.POST.getlist('reasons')
+        if not reasons:
```
taymonbeal: This should be part of form validation.

eahub/profiles/views.py
```diff
@@ -173,3 +180,38 @@ def delete_profile(request):
...
+                'profile_name': self.reportee.name,
+                'profile_url': "https://{0}/profile/{1}".format(get_current_site(request).domain,self.reportee.slug),
+                'reasons': ', '.join(reasons)
+            })
+            send_mail(subject, message, settings.DEFAULT_FROM_EMAIL, recipient_list=settings.LEAN_MANAGERS)
+            return redirect('/{0}/{1}/report-abuse-done'.format(self.type,self.reportee.slug))
+
+
+
+def report_abuse(request, slug):
```
taymonbeal: See above on `FormView`. Also, this code looks basically the same across both apps; maybe separate it out into a shared dependency?

eahub/localgroups/views.py
```diff
@@ -18,6 +18,8 @@
 from .forms import LocalGroupForm
 from .models import LocalGroup
 from ..profiles.models import Profile
+from ..profiles.views import ReportAbuse
```
taymonbeal: Seems like this is specific to neither profiles nor local groups, so should be in some common directory like `base` where they can both import it from.

eahub/templates/eahub/group.html
```diff
@@ -147,6 +147,14 @@ <h1 class="word-break">{{ group.name }}</h1>
...
+  <div class="text-center margin-top report-btn">
+    <div class="btn-group btn-group-sm text-center" role="group" id="report_profile_abuse_toggle">
+      <a href="{% url 'report_abuse_group' group.slug  %}"><button type="button" class="btn btn-default"><i class="fa fa-flag"></i> Report abuse</a>
+    </div>
+  </div>
+  
+
 </div>
-<script src="/static/scripts/group.js"></script>
+<script src="/static/scripts/report_buttons.js"></script>
```
taymonbeal: Use the `{% static %}` tag to link to static assets. It makes the site faster, because Whitenoise makes them cacheable.

eahub/base/static/scripts/report_buttons.js
sebgrebe: Right, this is no longer necessary - left from a previous implementation I abandoned. Will delete, thanks

eahub/localgroups/urls.py
sebgrebe: Yes, good idea.Redirect to the profile that is reported on and added flash message there.

eahub/localgroups/views.py
sebgrebe: implemented

eahub/profiles/views.py
sebgrebe: implemented

eahub/localgroups/views.py
sebgrebe: done

eahub/profiles/views.py
sebgrebe: put this into formview

eahub/base/forms.py
```diff
@@ -0,0 +1,14 @@
+from django import forms
+
+class ReportAbuseForm(forms.Form):
+    CHOICES = [('Spam', 'Spam'),
+               ('Fake account', 'Fake account'),
+               ('Offensive content', 'Offensive content')]
+    reasons = forms.MultipleChoiceField(choices=CHOICES, widget=forms.CheckboxSelectMultiple)
+
+    def clean(self):
```
taymonbeal: You don't need this. Fields in Django forms are required by default.

eahub/base/views.py
```diff
@@ -130,6 +140,48 @@ class RobotsTxtView(base.TemplateView):
     template_name = "robots.txt"
     content_type = "text/plain; charset=utf-8"
 
+class ReportAbuse:
```
taymonbeal: There doesn't seem to be any strong reason for this to be a separate class from the view.

eahub/base/views.py
```diff
@@ -130,6 +140,48 @@ class RobotsTxtView(base.TemplateView):
...
+    def __init__(self, reportee, type):
+        self.reportee = reportee
+        self.type = type
+
+    def send(self, request, form):
+        reasons = form.cleaned_data
+        subject = "EA {0} reported as abuse: {1}".format(self.type, self.reportee.name)
```
taymonbeal: These can be f-strings.

eahub/base/views.py
```diff
@@ -130,6 +140,48 @@ class RobotsTxtView(base.TemplateView):
...
+        message = render_to_string('emails/report_{}_abuse.txt'.format(self.type), {
+            'profile_name': self.reportee.name,
+            'profile_url': "https://{0}/profile/{1}".format(get_current_site(request).domain,self.reportee.slug),
+            'reasons': ', '.join(reasons)
+        })
+        send_mail(subject, message, settings.DEFAULT_FROM_EMAIL, recipient_list=settings.LEAN_MANAGERS)
+        messages.success(
+            request,
+            ''' Thank you, we have received your report. Our admin team will send you an email once they have looked into it. ''',
```
taymonbeal: No need for triple-quoted strings when it's already only one line.

eahub/base/views.py
```diff
@@ -130,6 +140,48 @@ class RobotsTxtView(base.TemplateView):
...
+        )
+        return redirect('/{0}/{1}'.format(self.type,self.reportee.slug))
```
taymonbeal: Again, use `get_absolute_url` here.

eahub/base/views.py
```diff
@@ -130,6 +140,48 @@ class RobotsTxtView(base.TemplateView):
...
+        return report_abuse.send(self.request, form)
+
+    def get_report_abuse(self):
+        return ReportAbuse(self.profile(), 'profile')
+
+    def profile(self):
+        return Profile.objects.get(slug=self.kwargs['slug'])
+
+
+class ReportGroupAbuseView(ReportProfileAbuseView):
```
taymonbeal: These two classes are siblings, so one shouldn't extend the other; they should both extend a common ancestor. That way, you can put the ancestor in `base` and the app-specific ones in those apps.

eahub/profiles/views.py
```diff
@@ -37,6 +38,7 @@ def profile_redirect_from_legacy_record(request, legacy_record):
     return redirect("profile", slug=profile.slug, permanent=True)
 
 
+
```
taymonbeal: ?

eahub/profiles/views.py
```diff
@@ -6,6 +6,7 @@
 from django.http import HttpResponse
 from django.shortcuts import render, redirect
 from django.contrib.auth.decorators import login_required
+from django.views.decorators.http import require_POST
```
taymonbeal: No longer needed.

eahub/base/views.py
```diff
@@ -130,6 +140,48 @@ class RobotsTxtView(base.TemplateView):
...
+    def __init__(self, reportee, type):
+        self.reportee = reportee
+        self.type = type
+
+    def send(self, request, form):
+        reasons = form.cleaned_data
+        subject = "EA {0} reported as abuse: {1}".format(self.type, self.reportee.name)
+        message = render_to_string('emails/report_{}_abuse.txt'.format(self.type), {
+            'profile_name': self.reportee.name,
+            'profile_url': "https://{0}/profile/{1}".format(get_current_site(request).domain,self.reportee.slug),
```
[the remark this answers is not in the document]
sebgrebe: This gets met http://localhost:8000/group/ea-klaipeda/report-abuse//group/ea-klaipeda/

eahub/base/views.py
sebgrebe: see comment above

PR #655, opened 2019-04-23
648 mailchimp signup

### LINKED ISSUE #648: Add Mailchimp signup form to site
https://github.com/rtcharity/eahub.org/issues/597#issuecomment-485268845

> How should the subscription flow work? The simplest option is to just embed a Mailchimp signup form on our site somewhere.

sebgrebe:
Fixes #648 

- Adds signup form for mailchimp newsletter
- Email required, first and last name optional. Or do we only want a "name" field as our user models don't distinguish between first and last names?
- We might want to add some text about the newsletter ("We'll send you monthly updates about the Hub blablabla"


eahub/templates/eahub/newsletter.html
```diff
@@ -0,0 +1,49 @@
...
+<div class="panel panel-default">
+<div class="panel-body">
+<div id="mc_embed_signup">
+<form action="https://effectivealtruismhub.us11.list-manage.com/subscribe/post?u=48139d2c8cadf4a1bce00bb6f&amp;id=7051ba2a99" method="post" id="mc-embedded-subscribe-form" name="mc-embedded-subscribe-form" class="validate" target="_blank" novalidate>
+<div id="mc_embed_signup_scroll">
+	<h2>Subscribe to our Newsletter</h2>
+  <div class="col-md-4 col-md-offset-4">
+  <div class="mc-field-group field">
+  	<label for="mce-EMAIL">Email Address  <span>*</span></label>
+  	<input type="email" value="" name="EMAIL" class="required email" id="mce-EMAIL">
```
taymonbeal: Can we prefill the fields if the user's logged in? I wouldn't ask if it was going to be hard, but I think it should be easy.

taymonbeal:
Yes, just have the one name unless Mailchimp makes this intractable. Let me know if you need help.

@michaltrzesimiech and @Richenda might have opinions on verbiage. Note that saying we'll email monthly might constrain us from emailing outside that schedule; not sure what the relevant regulations are. We also might want to make clear that this is the EA Hub newsletter; there's a separate LEAN newsletter.

taymonbeal:
Oh, and we need to link or embed this from elsewhere on the site so people can find it.

PR #669, opened 2019-04-29
Plato quick ux fixes

### LINKED ISSUE #505: Change "Local organiser of" to "Organiser of"
image

### LINKED ISSUE #552: Get rid of "on our website" from "This website uses cookies to ensure you get the best experience on our website."
image

image

### LINKED ISSUE #555: Soften the tone of "Your password was determined to have been involved in a major security breach."
When setting up my password to be "TestPassword", the message got me scared. My first thought was that I just attempted a major security breach on the Hub.

image

How about we soften that to "Your password is not secure enough for it was determined to have been involved in a major breach before."

sebgrebe:
closes #555 
closes #505 
closes #552


eahub/templates/base.html
```diff
@@ -33,24 +33,8 @@
...
-          "background": "#A7DBAB",
-          "text": "white"
-        }
-      },
-      "content": {
-        "href": "https://eahub.org/privacy-policy"
-      }
-    })});
-    </script>
+    <script src="static/scripts/cookieconsent/cookieconsent.min.js"></script>
```
taymonbeal: Use `{% static %}` to link to static assets.

eahub/config/settings.py
```diff
@@ -196,6 +196,7 @@
 
 # Django PWNED Passwords
 PWNED_VALIDATOR_FAIL_SAFE = False
+PWNED_VALIDATOR_ERROR = "Your password is not secure enough for it was determined to have been involved in a major breach before."
```
taymonbeal: This sentence seems garbled, somehow. Busy right now but will try to think of better wordings.

taymonbeal: Out of alphabetical order.

sebgrebe: Changed to "Your password is not secure enough". Maybe that's detailed enough?

taymonbeal: No, this is in a context where the user is already being told that and we're trying to explain why.

Honestly I feel like the original text was good enough here.

sebgrebe: I'm probably missing something, but in which sense was "the user already being told" that their password isn't secure enough? This message comes up when a user is trying to reset their password with an easily guessable password such as "TestPassword". As far as I can see they have not been told anything previous to that. And it seems that "Your password is not secure enough" is all the info the user needs to know. I agree with @michaltrzesimiech's original point that the original wording can be quite alarming and potentially misleading.

That said, if you feel strongly about this and @michaltrzesimiech on the other hand not too much, let's stick with the original wording.

taymonbeal: Ah, sorry, I wasn't clear. The issue is that in this context the user is shown a bulleted list of which specific password rules their password fails. "Not secure enough" doesn't make sense in that context.

sebgrebe: How about we leave it almost exactly as it was, but add "in the past": makes clear that the user has not caused a security breach:
"Your password was determined to have been involved in a major security breach in the past"

taymonbeal: Maybe we could do better than that by actually linking to the Pwned Passwords database so that the user has some clue of what's going on and where we got their old passwords from.

sebgrebe: add link to URL.

eahub/config/settings.py
```diff
@@ -195,8 +195,10 @@
 CRISPY_TEMPLATE_PACK = "bootstrap3"
 
 # Django PWNED Passwords
+PWNED_VALIDATOR_ERROR = "Your password was determined to have been involved in a major security breach in the past: https://haveibeenpwned.com/"
```
taymonbeal: Can you remove the URL from the text and instead make it into a link to https://haveibeenpwned.com/Passwords? HTML is allowed here.

sebgrebe: Tried this, HTML wasn't parsed but came out as strings

taymonbeal: Ah, sorry. Here's how to make it work: https://stackoverflow.com/questions/10706296/how-do-i-input-html-into-the-help-text-of-a-django-form-field

sebgrebe: Done, required though to break up the form

taymonbeal: I'm confused, I don't see this.

sebgrebe: sorry, push didn't go through. It's there now

eahub/templates/account/password_change.html
```diff
@@ -5,9 +5,16 @@
 {% block content %}
 <div class="container user-action">
 <h1>Change Password</h1>
-<form method="POST" action="{% url 'account_change_password' %}" class="password_change">
+<form method="POST" action="{% url 'account_change_password' %}">
       {% csrf_token %}
-      {{ form.as_p }}
+      {% for field in form %}
```
taymonbeal: There's no need to break up the form. You can use `mark_safe` in `settings.py` like in the Stack Overflow answer.

sebgrebe: Implemented.

eahub/config/settings.py
```diff
@@ -195,8 +197,10 @@
 CRISPY_TEMPLATE_PACK = "bootstrap3"
 
 # Django PWNED Passwords
+PWNED_VALIDATOR_ERROR = forms.ValidationError(mark_safe("Your password was determined to have been involved in a major security breach in the <a target='_blank' href='https://haveibeenpwned.com/passwords'>past</a>."))
 PWNED_VALIDATOR_FAIL_SAFE = False
 
+
```
taymonbeal: Extra newline.

eahub/config/settings.py
```diff
@@ -195,8 +197,10 @@
 CRISPY_TEMPLATE_PACK = "bootstrap3"
 
 # Django PWNED Passwords
+PWNED_VALIDATOR_ERROR = forms.ValidationError(mark_safe("Your password was determined to have been involved in a major security breach in the <a target='_blank' href='https://haveibeenpwned.com/passwords'>past</a>."))
```
taymonbeal: Do you need the `ValidationError` here? I think just `mark_safe` works.

sebgrebe: Done. Feel free to make any other changes you feel are necessary.

eahub/config/settings.py
sebgrebe: removed

eahub/config/settings.py
taymonbeal: I think this commit didn't get pushed.

PR #675, opened 2019-05-05
Filter out blank profiles and groups

### LINKED ISSUE #124: Sorting on lists puts blanks last

sebgrebe:
fixes #124, as explained in particular [here](https://github.com/rtcharity/eahub.org/issues/124#issuecomment-489456820)

### DIFF
--- a/eahub/base/views.py
+++ b/eahub/base/views.py
@@ -82,7 +82,7 @@ def groups(request):
     })
 
 def getGroupsData():
-    rows = Group.objects.all()
+    rows = Group.objects.exclude(name__isnull=True)
     map_data = [
         {
             "lat": x.lat,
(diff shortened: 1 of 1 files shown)


sebgrebe:
> `name` is a non-nullable column in both tables, so this change doesn't do anything. What did you mean for it to do?

Should be clear from the comment I linked. Let me know if you have more questions. The issue is that there are old profiles in which name is null

taymonbeal:
There are no profiles with null or blank names. Can you PM me a link to one that you think is like this?

sebgrebe:
Close PR as ticket is outdated

Issue #647, opened 2019-04-22
Order list of areas of expertise alphabetically


sebgrebe:
![Screen Shot 2019-04-22 at 19 59 11](https://user-images.githubusercontent.com/15908208/56519178-4fed3e00-6539-11e9-840a-1f213fb5ae01.png)

PR #679, opened 2019-05-07
Order list of areas of expertise alphabetically

michaltrzesimiech:
Closes https://github.com/rtcharity/eahub.org/issues/647


eahub/templates/account/signup.html
```diff
@@ -21,6 +21,8 @@
 
 {% block form_privacy_note %}
 
+<br>
```
taymonbeal: This seems to be left over from a different PR.

eahub/profiles/models.py
```diff
@@ -42,31 +42,31 @@ class CauseArea(enum.Enum):
...
-    OTHER_SCIENCE = 17
-    EVENT_PLANNING = 18
-    FINANCE = 19
-    GRAPHIC_DESIGN = 20
-    JOURNALISM = 21
-    LAW = 22
-    PHILANTHROPY_EARNING_TO_GIVE = 23
-    PUBLIC_SPEAKING = 24
-    RECRUITMENT = 25
+    COMMUNICATIONS = 1
```
taymonbeal: You can't change which name goes with which number. Those numbers are stored in the database.

michaltrzesimiech: Then, I understand, that I should add new records to the database and figure out how to sort this list by different means

Issue #573, opened 2019-04-08
Add new cause areas


michaltrzesimiech:
Rationality
Mental Health

taymonbeal:
Don't we already have climate change?

michaltrzesimiech:
We do. We missed Mental Health and Rationality

PR #692, opened 2019-05-13
New cause areas

michaltrzesimiech:
Closes https://github.com/rtcharity/eahub.org/issues/573

### DIFF
--- a/eahub/profiles/management/commands/profilesfromlegacyhub.py
+++ b/eahub/profiles/management/commands/profilesfromlegacyhub.py
@@ -19,6 +19,8 @@ def collect_cause_areas(
     cause_prioritisation,
     meta,
     climate_change,
+    mental_health,
+    rationality,
 ):
     cause_areas = []
     if global_poverty:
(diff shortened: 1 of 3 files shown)


eahub/profiles/management/commands/profilesfromlegacyhub.py
```diff
@@ -19,6 +19,8 @@ def collect_cause_areas(
     cause_prioritisation,
     meta,
     climate_change,
+    mental_health,
```
taymonbeal: Revert this; as written it'll cause a runtime error, and we're not planning on re-running this import anyway, I think.

eahub/profiles/models.py
```diff
@@ -37,36 +39,38 @@ class CauseArea(enum.Enum):
...
-    OTHER_SCIENCE = 17
-    EVENT_PLANNING = 18
-    FINANCE = 19
-    GRAPHIC_DESIGN = 20
-    JOURNALISM = 21
-    LAW = 22
-    PHILANTHROPY_EARNING_TO_GIVE = 23
-    PUBLIC_SPEAKING = 24
-    RECRUITMENT = 25
+    COMMUNICATIONS = 1
```
taymonbeal: This needs to be reverted too.

PR #693, opened 2019-05-13
Adding ALLFED

michaltrzesimiech:
Closes https://github.com/rtcharity/eahub.org/issues/592


eahub/templates/account/signup.html
```diff
@@ -25,4 +25,4 @@
   Read more about our approach to privacy in our <a href="{% url 'privacyPolicy' %}">Privacy Policy</a>.
 </div>
 
-{% endblock %}
+{% endblock %}
```
taymonbeal: You seem to have accidentally removed a trailing newline here. If your editor is doing this, fix its configuration.

eahub/profiles/models.py
```diff
@@ -75,7 +75,7 @@ class ExpertiseArea(enum.Enum):
         GOVERNMENT_AND_POLICY: "Government and policy",
         ENTREPRENEURSHIP: "Entrepreneurship",
         SOFTWARE_ENGINEERING: "Software engineering",
-        AI_TECHNICAL_EXPERTISE: "AI technical expertise",
+        SOFTWARE_ENGINEERING_AI: "AI technical expertise",
```
taymonbeal: Do you want to change the label of this since you're changing the name? Or is that not the idea?

Issue #700, opened 2019-05-13
Add a view for 410 error


michaltrzesimiech:

victor-yunenko:
Closing due to inactivity, please feel free to reopen if needed.

Issue #603, opened 2019-04-15
Put up 410 page for donations


taymonbeal:
This affects the following URLs:
- `/donations`
- `/donations/past/animal-welfare`
- `/donations/past`
- `/donations/past/add`
- `/donations/plans/global-poverty`
- `/donations/plans/animal-welfare`
- `/donations/plans/all/alphabetical`
- `/donations/add`
- `/node/9`

patbl:
What's the advantage of a 410 over a 404?

victor-yunenko:
Closing due to inactivity, please feel free to reopen if needed.

PR #701, opened 2019-05-13
Adding a handler for 410

michaltrzesimiech:
Closes https://github.com/rtcharity/eahub.org/issues/700

### DIFF
--- a/eahub/base/views.py
+++ b/eahub/base/views.py
@@ -172,7 +172,6 @@ class ReportAbuseView(FormView):
 def healthCheck(request):
     return HttpResponse(status=204)
 
-
 def page_not_found(request, exception):
     if not isinstance(exception, exceptions.Quiet404):
         mail.mail_managers(
@@ -180,3 +179,6 @@ def page_not_found(request, exception):
             loader.render_to_string("emails/broken_link.txt", {"request": request}),
         )
     return defaults.page_not_found(request, exception)
+
+def gone(request):
+    return HttpResponse(status=410)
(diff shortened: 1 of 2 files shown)


eahub/config/urls.py
```diff
@@ -52,3 +52,4 @@
 ]
 
 handler404 = "eahub.base.views.page_not_found"
+handler410 = "eahub.base.views.gone"
```
taymonbeal: This doesn't do anything; these are exception handlers and there are no exceptions in Django that result in 410 responses.

eahub/base/views.py
```diff
@@ -172,11 +172,13 @@ def form_valid(self, form):
...
 def page_not_found(request, exception):
     if not isinstance(exception, exceptions.Quiet404):
         mail.mail_managers(
             "Broken link",
             loader.render_to_string("emails/broken_link.txt", {"request": request}),
         )
     return defaults.page_not_found(request, exception)
+
+def gone(request):
+    return HttpResponse(status=410)
```
taymonbeal: This just returns a blank page to the user, which I don't think we want. They should see a page built from our templates, with an explanation of why the specific thing they were looking for doesn't exist anymore.

Also, your editor still isn't handling end-of-file newlines properly.

eahub/config/urls.py
michaltrzesimiech: I wanted to split this in 2 moves, but you're right. I'll just apply this to https://github.com/rtcharity/eahub.org/issues/603

taymonbeal: No, I mean there's no way this can ever do anything, because there's no such thing in Django as a 410 exception and you can't add one. Which exceptions return which status codes is hardcoded into the framework.

Issue #604, opened 2019-04-15
Handle legacy group URLs


taymonbeal:
This affects the following URLs:
- `/groups/<slug>`

taymonbeal:
Note that this doesn't fully resolve the issue, because many groups are using different slugs now than they were on the old Hub, so those links are still broken. To fully resolve this we need groups to be able to have aliases, like how profiles do today, and we need to import all the old slugs as aliases.

victor-yunenko:
I believe we have another ticket for that which would allow url editing, I'll close this issue for now then.

PR #702, opened 2019-05-14
Handling legacy group requests

michaltrzesimiech:
Closes https://github.com/rtcharity/eahub.org/issues/604


eahub/config/urls.py
```diff
@@ -36,8 +36,9 @@
     path('groups/get-a-website/', views.LegacyRedirectView.as_view(url='https://github.com/rtcharity/lean-site-template')),
     path('groups/local-group-support-overview-lean-cea-and-eaf/', views.LegacyRedirectView.as_view(url='https://forum.effectivealtruism.org/posts/Cvi7hnTYMk5qutkDg/local-effective-altruism-network-s-new-focus-for-2019')),
     path('groups/resources/giving-games/', views.LegacyRedirectView.as_view(url='https://docs.google.com/document/d/1g5G0PvYFs7cAbAZ8ANI_wnsv7DjtBmq4_SQcg44-HDA/edit')),
+    path('groups/<slug:slug>/', views.LegacyRedirectView.as_view(pattern_name='groups')),
```
taymonbeal: This doesn't look right; I'd expect it to be `pattern_name="group"`. Have you tested it locally?

michaltrzesimiech: You're right. 

> Have you tested it locally?

I wasn't able to set up the local env. Raised an issue here: https://github.com/rtcharity/eahub.org/issues/709

michaltrzesimiech: I tested it now and it looks like we want `groups`. `localhost:8000/group` returns 404.

taymonbeal: The pattern you want to be pointing at is https://github.com/rtcharity/eahub.org/blob/df156c575b7ebfb74c1537ca82f228f9c04b8ac5/eahub/localgroups/urls.py#L10. `/group` is indeed not a URL that we have, but I don't see what that has to do with anything; it accepts a pattern name, not a URL path. Did you not get a 500 error with `pattern_name='groups'`?

michaltrzesimiech: Sorry, I got confused. Obviously, you're right. I tested it and it works as it' supposed.

Issue #520, opened 2019-03-29
Recreate /newsletter


michaltrzesimiech:
Based on 

> Referrer: https://www.effectivealtruism.org/get-involved/apply-for-an-ea-job/ Requested URL: /newsletter User agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_7_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/49.0.2623.112 Safari/537.36

Snapshot: http://web.archive.org/web/20170214024606/https://eahub.org/newsletter

PR #703, opened 2019-05-14
Recreate /newsletter

michaltrzesimiech:
Closes https://github.com/rtcharity/eahub.org/issues/520

### DIFF
--- a/eahub/config/urls.py
+++ b/eahub/config/urls.py
@@ -21,6 +21,7 @@ urlpatterns = [
     path('groups/', views.groups, name='groups'),
     path('admin/', admin.site.urls, name='admin'),
     path('about/', views.about, name='about'),
+    path('newsletter/', TemplateView.as_view(template_name='eahub/newsletter.html'), name='newsletter'),
     path('about/newsletter/', TemplateView.as_view(template_name='eahub/newsletter.html'), name='newsletter'),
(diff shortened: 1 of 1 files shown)


eahub/config/urls.py
```diff
@@ -21,6 +21,7 @@
     path('groups/', views.groups, name='groups'),
     path('admin/', admin.site.urls, name='admin'),
     path('about/', views.about, name='about'),
+    path('newsletter/', TemplateView.as_view(template_name='eahub/newsletter.html'), name='newsletter'),
```
taymonbeal: We shouldn't have the same content at two different URLs (and we _can't_ have two URL patterns in Django with the same name; that will break things). One of these should redirect to the other.

taymonbeal: Did you mean for one of these to be a redirect?

michaltrzesimiech: Yeah, I made a dumb guess. I've read In here on how to do this properly (I think): https://stackoverflow.com/questions/42614172/how-can-i-redirect-from-view-in-django/42614265, but I guess I'll hold on until I have a local environment running.

taymonbeal: Use the `LegacyRedirectView`; we've got a bunch of these elsewhere in the file.

michaltrzesimiech: Lol, you're right. Done that.

Issue #704, opened 2019-05-14
Integrate Google AdSense


michaltrzesimiech:
Also needed for https://github.com/rtcharity/eahub.org/issues/614

michaltrzesimiech:
We decided to not integrate AdSense only to get ads.txt. We probably don't want ads.txt anyway (https://github.com/rtcharity/eahub.org/issues/614).

PR #705, opened 2019-05-14
Integrate AdSense

michaltrzesimiech:
Closes https://github.com/rtcharity/eahub.org/issues/704

### DIFF
--- a/eahub/templates/base.html
+++ b/eahub/templates/base.html
@@ -41,6 +41,13 @@
     }(window,document,"Smooch","5caa256dfea08a001172c8ea");
     </script>
     <script>Smooch.init({ appId: '5caa256dfea08a001172c8ea' });</script>
+    <script async src="//pagead2.googlesyndication.com/pagead/js/adsbygoogle.js"></script>
+    <script>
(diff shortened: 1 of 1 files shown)


michaltrzesimiech:
We don't, this step is required to generate ads.txt

michaltrzesimiech:
Long story short, we generate that in Ad Manager, but `To sign up for Google Ad Manager, you must first have a Google AdSense account.`. To activate AdSense account, we need to include this snippet in our <head>.

https://support.google.com/admanager/answer/7544382

michaltrzesimiech:
Or, we could maybe just use a random ads.txt

michaltrzesimiech:
Curiously, we no longer have an option to enable the thing we don't want in the first place, because they discovered one of these issues https://support.google.com/adsense/answer/1348688?hl=en and we should get rid of it.

michaltrzesimiech:
I currently opt for finding another way to implement ads.txt.

taymonbeal:
We don't need an AdSense account, it was never the plan to sign up for one. I'll explain in more detail later.

Issue #612, opened 2019-04-15
Figure out what /user/profiles/offers did


taymonbeal:
This affects the following URLs:
- `/user/profiles/offers`

victor-yunenko:
Closing since I'm assuming that the old links have been updated/dropped during the last year. I also wasn't able to find that url in google index. Please feel free to reopen if needed.

PR #708, opened 2019-05-14
Redirect user/profiles/offers to profiles

michaltrzesimiech:
Closes https://github.com/rtcharity/eahub.org/issues/612

### DIFF
--- a/eahub/config/urls.py
+++ b/eahub/config/urls.py
@@ -47,7 +47,7 @@ urlpatterns = [
     path('user/<int:legacy_record>/', views.LegacyRedirectView.as_view(pattern_name='profile_legacy')),
     path('user/<slug:slug>/', views.LegacyRedirectView.as_view(pattern_name='profile')),
     path('user/profiles/', views.LegacyRedirectView.as_view(pattern_name='profiles')),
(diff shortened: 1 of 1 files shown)


eahub/config/urls.py
```diff
@@ -47,6 +47,7 @@
     path('user/<int:legacy_record>/', views.LegacyRedirectView.as_view(pattern_name='profile_legacy')),
     path('user/<slug:slug>/', views.LegacyRedirectView.as_view(pattern_name='profile')),
     path('user/profiles/', views.LegacyRedirectView.as_view(pattern_name='profiles')),
+    path('user/profiles/offers', views.LegacyRedirectView.as_view(pattern_name='profiles')),
```
taymonbeal: Trailing slash needed.

michaltrzesimiech: > Do you happen to know what exactly was at this URL on the old Hub?

It wasn't cached or documented anywhere. I doubt that it was in regular use anyway. My opinion is that spending more time on figuring this out isn't worth it.

taymonbeal:
I'm closing this for now because I want to come back to it later, when more immediate priorities have quieted, and actually open up the old Hub and systematically figure out what all these things did. Feel free to reopen if you really want to do something.

Issue #613, opened 2019-04-15
Figure out what /map/people/only-on-map did


taymonbeal:
This affects the following URLs:
- `/map/people/only-on-map`

PR #710, opened 2019-05-14
Directing 'map/people/only-on-map' to

michaltrzesimiech:
Closes https://github.com/rtcharity/eahub.org/issues/613


eahub/config/urls.py
```diff
@@ -41,6 +41,7 @@
     path('links/', views.LegacyRedirectView.as_view(url='https://resources.eahub.org (temporarily)')),
     path('map/', views.LegacyRedirectView.as_view(pattern_name='profiles')),
     path('map/people/all/', views.LegacyRedirectView.as_view(pattern_name='profiles')),
+    path('map/people/only-on-map', views.LegacyRedirectView.as_view(pattern_name='profiles')),
```
taymonbeal: Needs a trailing slash.

michaltrzesimiech: Done. Thanks!

taymonbeal:
Still wondering about above question.

michaltrzesimiech:
Sorry, missed that! I'm not convinced it's useful at all, but I imagine we could change our minds and so the setup in the future?

PR #786, opened 2019-06-20
/groups lists groups by distance from user if user is logged in and has location data

david-mears:


eahub/base/views.py
```diff
@@ -94,28 +94,77 @@ def groups(request):
     )
 
 
-def get_groups_data():
-    rows = Group.objects.all()
+def get_groups_data(user, by_distance=False):
```
taymonbeal: Boolean parameters are often a code smell. Could this make sense as two separate functions?

eahub/base/views.py
```diff
@@ -94,28 +94,77 @@ def groups(request):
...
+    if hasattr(user, "profile") and by_distance:
```
taymonbeal: I think it's a bit more performant/idiomatic to just assume that `user.profile` exists and then catch the `Profile.DoesNotExist` exception in case it doesn't. (Almost all users have profiles.)

eahub/base/views.py
```diff
@@ -94,28 +94,77 @@ def groups(request):
...
+        if hasattr(user.profile, "lat"):
```
taymonbeal: This check doesn't work; all profiles have a `lat` attribute. You want `if user.profile.lat is not None`.

Also, what happens if the user has a profile but it doesn't have a location? I think this would crash in that case.

Also also, I think you should check for both `lat` and `lon`; normally they should go together, but this isn't strictly enforced. (I mean, it is on users, but not on staff.)

eahub/base/views.py
```diff
@@ -94,28 +94,77 @@ def groups(request):
...
-        for x in rows
-        if x.lat and x.lon
+        for group in rows
+        if group.lat and group.lon
     ]
     return {"rows": rows, "map_data": map_data}
 
 
+def get_groups_by_distance(profile):
+    # Difference in latitude (multiplied by 2 because latitude ranges only to 90)
```
taymonbeal: The correct way to calculate distances across latitude-and-longitude points is the haversine formula. Contrary to the below comment, you _can_ in fact implement this in a Django lookup; Django offers bindings to database math functions, including square-root and trigonometric functions.

eahub/base/views.py
```diff
@@ -94,28 +94,77 @@ def groups(request):
...
-        for x in rows
-        if x.lat and x.lon
+        {
+            "lat": profile.lat,
+            "lng": profile.lon,
+            "label": profile.name,
+            "path": f"/profile/{profile.slug}",
+        }
+        for profile in rows
+        if profile.lat and profile.lon
```
taymonbeal: `if profile.lat is not None and profile.lon is not None`. The implicit coercion returns false if the latitude or longitude happens to be exactly 0.0, which is unlikely but possible.

eahub/base/views.py
```diff
@@ -94,28 +94,77 @@ def groups(request):
...
     rows = Profile.objects.visible_to_user(user)
     map_data = [
-        {"lat": x.lat, "lng": x.lon, "label": x.name, "path": f"/profile/{x.slug}"}
-        for x in rows
-        if x.lat and x.lon
+        {
+            "lat": profile.lat,
+            "lng": profile.lon,
+            "label": profile.name,
+            "path": f"/profile/{profile.slug}",
```
taymonbeal: This should be using Django's `reverse` function instead of manually constructing the URL path.

taymonbeal:
Also, I'd really rather have unit tests for all this, especially since they'd have caught some of the bugs. If you wanted to add some I'd be happy to help.

PR #798, opened 2019-06-28
Add privacy setting for local groups.

### LINKED ISSUE #776: Add a tickbox that hides a group profile from UI
Some organizers don't want to list their groups on the Hub and so we would like a way to hide those groups from the UI (but keep in the database).

mkbehr:
Add an is_public field to local groups: if it's set to false, the group won't be visible except to its organisers.

Closes #776

### DIFF
--- a/eahub/base/views.py
+++ b/eahub/base/views.py
@@ -37,7 +37,7 @@ class CustomisedPasswordChangeView(PasswordChangeView):
 
 
 def index(request):
-    groups_data = get_groups_data()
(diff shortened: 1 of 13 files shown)


mkbehr:
I tried to hit all the cases that might show a group, but I didn't find a way to comprehensively test that, so I'd appreciate a second pair of eyes there.

taymonbeal:
We need to figure out how membership in private groups should work. @michaltrzesimiech, do you have any thoughts on this? Should private groups have members at all?

Issue #598, opened 2019-04-14
Add a LinkedIn link to profile


michaltrzesimiech:

Richenda:
I would see this as very low priority

Richenda:
I'm worried it would be a lot of work for very little impact

michaltrzesimiech:
I was thinking of adding an optional link to LinkedIn, which isn't much work at all. Also, I see this as potentially very impactful in terms of helping hiring managers use the Hub to find talent.

sebgrebe:
Seconding Michael here, it would be quite good to help EA orgs hiring to find the Linkedin links of members quickly and easily and it is easy to implement

michaltrzesimiech:
Yeah, we should definitely do that. How about we add (to profile) another panel like `Links` and include (optional) links to Facebook and to LinkedIn, and a personal website?

taymonbeal:
Did we decide anything about #337?

Richenda:
> Did we decide anything about #337?

Just that it isn't an immediate priority

michaltrzesimiech:
We should up the priority of this based on multiple prompts, including from organisational users

PR #818, opened 2019-08-18
Profiles: Add LinkedIn, Facebook, and personal-website links

patbl:
See #598.

I stuck these in the Edit Profile form because it was easy, and the URL fields seemed similar to the other form fields ("basic info").

image

I don't know what a good way of displaying the info is. I stuck some Font Awesome icons under the name, but obviously that's not great.

image

I didn't try to make this stuff look good because I wanted your input before I spent too much time on it.

It might be a good idea to split the migrations into a separate PR so there's no downtime, though I don't know whether enough people use this site for us to worry about a minute or so of downtime.


eahub/templates/eahub/profile.html
```diff
@@ -22,6 +22,15 @@
     </div>
     <div class="col-xs-12 col-md-6">
       <h1 class="word-break">{{ profile.name }}</h1>
+      {% if profile.linkedin_url %}
+      <a href="{{ profile.linkedin_url }}"><i class="fab fa-linkedin-in"></i></a>
```
sebgrebe: This doesn't act quite as I would expect. If I enter "www.linkedin.com" it creates the link eahub.org/seb/profile/www.linkedin.com. Only if you add "https://" does it create the expected URL

eahub/templates/eahub/profile.html
```diff
@@ -22,6 +22,15 @@
     </div>
     <div class="col-xs-12 col-md-6">
       <h1 class="word-break">{{ profile.name }}</h1>
+      {% if profile.linkedin_url %}
```
sebgrebe: As you said on slack, positioning it here is not ideal. 
Maybe we could place it in the summary box? Or we add an extra box under the summary box for URLs?

eahub/profiles/models.py
```diff
@@ -254,6 +254,9 @@ class Profile(models.Model):
     )
     city_or_town = models.CharField(max_length=100, blank=True)
     country = models.CharField(max_length=100, blank=True)
+    linkedin_url = models.CharField(max_length=400, blank=True)
```
sebgrebe: I think we should have some more validation here, i.e., whether the input is of a valid URL format. Django appears to have an [built-in URL class](https://docs.djangoproject.com/en/2.2/ref/models/fields/#urlfield), maybe that would be a good choice here

eahub/templates/eahub/edit_profile.html
```diff
@@ -44,6 +44,10 @@
 
 {{ form.summary|as_crispy_field }}
 
+{{ form.linkedin_url|as_crispy_field }}
```
sebgrebe: If you implement URL validation (see comment above), make sure that appropriate error messages appear here when a user enters an invalid input

eahub/templates/eahub/profile.html
patbl: If I use `URLField` and enter `www.linkedin.com` or `linkedin.com`, Django adds `http://` to the beginning, so this isn't a problem.

eahub/templates/eahub/edit_profile.html
patbl: This works automatically when I use `URLField`:

image

patbl:
How about something like this?

image

eahub/templates/eahub/profile.html
michaltrzesimiech: My first preference is if there was another panel below Summary called Links. Both of these icons could be placed there. If this was troublesome for any reason, the current position could work if there was a <br> to visually push the other row of buttons underneath. We should also use this color for the icons

image

Because we don't currently offer messaging (and "soon" isn't this year anymore), we could also use the purple buttons to lead to social platforms the links to which people could upload in their profiles.
 
image

So that when I'm visiting my own profile, the buttons say i.e. "Update LinkedIn URL", "Update Facebook URL". When I'm visiting someone else's profile, they say "LinkedIn", "Facebook", etc.

michaltrzesimiech:
> How about something like this?
> 
> image

Looks good! We should use this color to make it blend in. 

image

Let me know if you need help with changing it.

patbl:
image

Issue #730, opened 2019-05-28
Remove the duplicated "Log in" from the burger menu


michaltrzesimiech:
![image](https://user-images.githubusercontent.com/14195560/58490178-f649e600-816c-11e9-8b00-2bd3e0768948.png)

bionfinc:
@michaltrzesimiech should the "Sign up" field be removed from this hamburger menu as well?

PR #825, opened 2019-09-05
Remove "Log in" from burger menu

bionfinc:
Fixes: #730


bionfinc:
@taymonbeal Yes as I understood it the ticket was created because there was no need to have "Log in" in that menu since it was already in the nav. I had initially assumed there was a separate ticket for removing the "Sign up" link, but I added a commit to this pull request to remove "Sign up" as well.

sebgrebe:
closed because requested changes were not implemented within 10 months

PR #829, opened 2019-10-04
Add some unit tests

david-mears:
I've started on some unit tests. Here's my first small PR, would appreciate a review.

### DIFF
--- a/eahub/base/tests/test_base_models.py
+++ b/eahub/base/tests/test_base_models.py
@@ -1,0 +1,15 @@
+from django.test import TestCase
+from model_mommy import mommy
+
+
+class UserTests(TestCase):
+    def test_has_profile(self):
+        user = mommy.make("base.User")
+
+        # With no profile
+        self.assertFalse(user.has_profile())
+
+        # With a profile
+        mommy.make("Profile", user=user, slug="slug")
+
+        self.assertTrue(user.has_profile())
--- a/eahub/profiles/tests/test_profiles_models.py
+++ b/eahub/profiles/tests/test_profiles_models.py
@@ -11,7 +11,7 @@ class ProfileTests(TestCase):
     def test_profile_creation(self):
         profile = mommy.make("Profile", slug="slug", name="Derek Parfit")
         self.assertTrue(isinstance(profile, Profile))
-        self.assertTrue(str(profile), "Derek Parfit")
+        self.assertEqual(str(profile), "Derek Parfit")
 
     def test_geocode(self):
         profile = mommy.make(
--- a/eahub/profiles/tests/test_profiles_models.py
+++ b/eahub/profiles/tests/test_profiles_models.py
@@ -22,8 +22,8 @@ class ProfileTests(TestCase):
             country="UK",
         )
         profile.geocode()
-        self.assertAlmostEqual(profile.lat, 51.4893335)
-        self.assertAlmostEqual(profile.lon, -0.144055084527687)
+        self.assertTrue(51.2 <= profile.lat <= 51.8)
+        self.assertTrue(0.3 >= profile.lon >= -0.3)
 
     def test_has_cause_area_details(self):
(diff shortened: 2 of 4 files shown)


eahub/base/tests/test_base_models.py
```diff
@@ -0,0 +1,15 @@
...
+
+class UserTests(TestCase):
+    def test_has_profile(self):
+        user = mommy.make("base.User")
+
+        # With no profile
+        self.assertFalse(user.has_profile())
+
+        # With a profile
+        mommy.make("Profile", user=user, slug="slug")
```
david-mears: I worry that this isn't a good unit test because there's a dependency on another class (Profile) here. But I don't know of another way to test the behaviour of the method under test here, than to use a real Profile object. In other words, I'm not sure if in this case the most isolated type of test possible is an integration test with such a dependency.

david-mears: (Question still applies, but commit has moved after a git rebase.)

taymonbeal: This is a common misconception; it's totally fine to use real dependencies in unit tests, unless there's a specific reason why doing so inhibits hermeticity or is otherwise problematic. (There's some really good documentation about this at my workplace that I'm trying to get open sourced.)

eahub/base/tests/test_base_models.py
```diff
@@ -0,0 +1,15 @@
...
+from model_mommy import mommy
+
+
+class UserTests(TestCase):
+    def test_has_profile(self):
+        user = mommy.make("base.User")
+
+        # With no profile
+        self.assertFalse(user.has_profile())
+
```
taymonbeal: It'd be better to split this up into two separate test cases.

eahub/profiles/tests/test_profiles_models.py
```diff
@@ -0,0 +1,35 @@
...
+from model_mommy import mommy
+
+from eahub.profiles.models import Profile
+
+
+class ProfileTests(TestCase):
+    def setUp(self):
+        self.name = "Test Person"
+
+    def test_profile_creation(self):
```
taymonbeal: This is effectively testing the behavior of Model Mommy; I don't think it adds any coverage to our code.

eahub/profiles/tests/test_profiles_models.py
```diff
@@ -0,0 +1,35 @@
...
+            country="UK",
+        )
+        profile.geocode()
+        self.assertTrue(51.2 <= profile.lat <= 51.8)
+        self.assertTrue(0.3 >= profile.lon >= -0.3)
+
+    def test_has_cause_area_details(self):
+        profile_without_cause_area = mommy.make("Profile", slug="slug", name=self.name)
+        self.assertFalse(profile_without_cause_area.has_cause_area_details())
+
```
taymonbeal: Again, split this.

requirements.txt
```diff
@@ -41,3 +41,4 @@ pep8-naming==0.8.2
 isort[requirements]==4.3.20
 flake8-isort==2.7.0
 django-webpack-loader==0.6.0
+model-mommy==1.6.0
```
taymonbeal: I kind of want to use Factory Boy rather than Model Mommy because it already has pytest integrations. Do you think there's a good reason to go with Model Mommy?

eahub/base/tests/test_base_models.py
david-mears: Perhaps you can point me to some docs that exist outside your company @taymonbeal ?

taymonbeal: I'll try to find some. In the meantime, the really short version is, you don't have to swap out dependencies of the code under test; it's fine to use the real dependencies unless there's a specific technical reason why this is impossible or undesirable.

sebgrebe:
closed because requested changes were not implemented within 9 months

PR #846, opened 2019-11-06
README: Add instructions for using the Python debugger

patbl:


README.md
```diff
@@ -42,6 +42,31 @@ If the database schema has changed since last time, you'll need to run
...
+To use `pdb`, the Python debugger, follow these steps:
+
+1. Open `docker-compose.yml`. Under the `web` key, insert `stdin_open: true` and `tty: true`. It should look like this:
+
+```yaml
+services:
+  web:
+    stdin_open: true
+    tty: true
+```
```
patbl: Could we add this permanently? That would make using the debugger a bit easier. I'm not familiar with Docker, so I don't know offhand if that would be OK. I could look into it.

hnykda: Yeah, that's fine IMO

hnykda:
The way I enable attaching to debugger is to run the server as `docker-compose run --service-ports app` - that way the server starts and when the execution encounters `breakpoint()`, it drops in the debugger in the same window. Isn't that a bit easier as you don't have to attach and stuff?

What you did in the docker-compose is IMO also fine.

README.md
```diff
@@ -47,6 +47,34 @@ To see live changes to the frontend while developing,
...
+
+2. Run `docker-compose up`.
+3. Run `docker ps`. Find the value in the "CONTAINER ID" column that corresponds to the `eahuborg_web` image.
+4. Using the ID you found in the previous step, run `docker attach <container ID>`. The command should be something like `docker attach 362618269c67`.
+5. Insert `import ipdb; ipdb.set_trace()` wherever you need a debugger statement.
```
hnykda, line 66: I personally prefer to setting up the env variable `PYTHONBREAKPOINT=ipdb.set_trace` and then just use `breakpoint()` in the code

hnykda:
Ah, I just noticed that I commented on something a year old :facepalm: . How did I even got here? :smile:

victor-yunenko:
Wouldn't PyCharm open source debugger with its docker integration be easier & more powerful?

sebgrebe:
> 
> 
> Wouldn't PyCharm open source debugger with its docker integration be easier & more powerful?

I believe we tried this together and couldn't get it working, so at least in this case it wasn't easier

victor-yunenko:
Sure, sounds good to me 👍 

FYI I also enabled the setup that allows to run it without docker, there the debugger is of course working as usually. I can document it once we merge in the latest changes.

sebgrebe:
Closing this and implementing @hnykda's suggestion in new PR: #1062

Issue #791, opened 2019-06-28
Clearer error message for Pwned Passwords


taymonbeal:
People are still finding this confusing at least sometimes.

patbl:
This is what it looks like:

![image](https://user-images.githubusercontent.com/1735266/69489487-6d66eb00-0e2d-11ea-946d-8a80123c0ec7.png)

It's set here:

https://github.com/rtcharity/eahub.org/blob/e865ba778cbab10d651378a798bce6de2f3f254e/eahub/config/settings.py#L238-L241

PR #849, opened 2019-11-15
Updating the error message for pwned passwords

michaltrzesimiech:


david-mears:
Meta: when you make a PR, can you include at least a brief explanation of the reason or the context for the PR?

In terms of readability, I would suggest moving the link (I can't update this branch):
```python
"For your security, consider using a password that wasn't involved in "
"<a target='_blank' href='https://haveibeenpwned.com/passwords'>a security breach</a> "
"before."
```

In terms of language, I prefer "hasn't been involved" to "wasn't involved".

michaltrzesimiech:
Thanks, David. Introduced the changes you suggested. The context is: https://github.com/rtcharity/eahub.org/issues/791 (the current message is confusing to some of our users)

PR #857, opened 2019-12-05
Update contact email address

### LINKED ISSUE #843: Change mentions of lean at eahub.org to contact at eahub.org

patbl:
This changes lean@eahub.org to contact@eahub.org.

Fixes #843.


eahub/profiles/management/commands/sendreactivationemails.py
```diff
@@ -194,7 +194,7 @@ def handle(self, *args, **options):
             message = mail.EmailMultiAlternatives(
                 subject="Announcing the EA Hub 2.0",
                 body=loader.render_to_string("emails/reactivate.txt", context),
-                from_email="Michael from LEAN <lean@eahub.org>",
+                from_email="Michael from LEAN <contact@eahub.org>",
```
sebgrebe: I'm not sure this is covered by the ticket as I assume "mention" refers to things visible on the webpage. This probably doesn't matter much anyway since those reactivation emails have already been sent, but I'd rather leave this as it was before

sebgrebe: Actually, thinking about it, this should be changed here as well. You were right.

PR #867, opened 2020-02-22
allow null values for open_to_couchsurfers field

sebgrebe:
Fixes a bug where on profile creation, error `null value in column "open_to_couchsurfers" violates not-null constraint` occurs.


sebgrebe:
unnecessary commit - no actual bug

patbl:
Huh, I thought this was an actual bug—I got an error when I tried to sign up in development.

This fixed it: https://github.com/rtcharity/eahub.org/pull/883

PR #884, opened 2020-03-19
Allow users to make their email address visible to other users

### LINKED ISSUE #586: Add messaging

patbl:
Fixes #586.

I put the setting here:

image

It might make more sense to have it on the "Change Email Address" page, but I'm worried that people wouldn't think to look there because that's such an infrequent action for a user to take (my guess is that more than 90% of people who use the site wouldn't visit that page, unless it's involved in some confirmation step).

If you opt in, this button will appear to logged-in viewers of your profile:

image

It's a mailto link, so it requires that you have that configured correctly in your browser. I'm open to a design that in addition shows the actual email address, but that would be more work and I don't have a good idea of what it would look like offhand.


cafelow:
Just to clarify  - will the email address be displayed or is the idea that there will be a message button with the mailto link. 
I would greatly prefer the former unless it changes the security somehow. I've had feedback from groups organisers that clicking the "message" button on the group pages is really annoying because a lot of people don't set up their browser to process mailto links in a way that makes sense to them.

patbl:
> will the email address be displayed or is the idea that there will be a message button with the mailto link.

In the current implementation, the email address won't be displayed; there would only be the mailto link. I agree that this should be changed if many people don't have their browsers set up to handle mailto links.

There's not an obvious design that occurs to me. Do you have any ideas? If not, I could think about it a bit and post some ideas.

patbl:
@sebgrebe Good catch. I was checking only logged-in status—I totally forgot to use the setting that I went to all the trouble of adding lol.

I changed it from a button to a link that displays the email address because of @cafelow's concern about mailto links (it's still a mailto link, but you can copy the email address now):

image

One remaining question is whether a user who has made their email address visible should see it on their profile. Before when the button said "Email Joe Smith" it seemed wrong to include it, since you wouldn't email yourself. But now that it's more informational ("Email: joesmith@gmail.com"), it could be OK. Here's how it would look:

image

Does anyone have a preference?

sebgrebe:
@patbl Works, thanks for the fit.

> Does anyone have a preference?
I'd prefer if users would see their own email address. The page should be as close to what others see as possible.

sebgrebe:
In the comms email regarding new changes to be sent out soon, we should add that users need to make their email address public if they want to be contactable through the Hub

patbl:
image

cafelow:
Looks great Patrick. Yeah I think it is good to have the email address visible to others and to themselves when editing.

PR #886, opened 2020-03-29
Add fields for other_website and other_info to group

### LINKED ISSUE #873: Add optional "Other website" field to /group/*
Specified same as the "Website" field; the field is invisible if value is null

### LINKED ISSUE #874: Add optional "Other information" field to /group/*
Text, 5000 characters max, placed at the bottom, the field is invisible if value is null

sebgrebe:
closes #873 
closes #874. Unlike specified in the issue, I set max characters to 2000, as this is the same max characters as the analogous summary filed for profiles. I hope that was OK


sebgrebe:
@cafelow I was choosing between 2 different ways of where to place this other information field on the page. I went with option 1). Would you prefer 2)?

Option 1)
option1

Option 2)
option2

michaltrzesimiech:
Option 2 is better imo

PR #889, opened 2020-03-31
WIP: new webpack setup, without removed png images and html fixes for now

victor-yunenko:
Todo
- extend the list of required favicons in `eahub/base/static/vendor/main.js`
- add DJANGO_ENV=prod env variable to the main server
- add more missing js and css addtoblock to statements as in `eahub/templates/eahub/index.html`
- convert the remaining images to required size and jpg


sebgrebe:
@victor-yunenko When running `docker-compose up`, I'm getting
"ModuleNotFoundError: No module named 'sekizai'"

requirements.txt
```diff
@@ -1,44 +1,131 @@
...
-django-feature-policy==2.3.0
-black==19.3b0
-pytest-black==0.3.7
-pytest-flake8==1.0.4
-pep8-naming==0.8.2
-isort[requirements]==4.3.20
-flake8-isort==2.7.0
-django-webpack-loader==0.6.0
-django-admin-relation-links==0.2.4
+#
```
sebgrebe, line 1: Why would you like to implement this change? Is this necessary for sorting the webpack stuff?
If not, I'd suggest to keep the PR as small as possible to achieve its main goal.

victor-yunenko: Hi @sebgrebe, regarding the necessity - I would say yes and no. In general I wouldn't advice to run a production server without compiled requirements, it isn't safe for a number of reason - I wouldn't like to describe them in text, but we can talk about it if you wish :)

sebgrebe: Let's do that! But let's make this change part of a different PR, to keep PRs manageable

victor-yunenko:
> It's not running at the moment, as when running docker-compose up, I'm getting
ModuleNotFoundError: No module named 'sekizai'

Did you happen to rebuild the docker? It's required after every change to requirements.txt

> It has broken the frontend tests. I'm getting

It sure did :)

Sorry probably I should have written in text, one of the reasons why I would like to get your help is:

> add more missing js and css addtoblock to statements as in eahub/templates/eahub/index.html

Basically right now most of the scripts won't work because they should be placed in correct html files, on the matter of what scripts must be placed where I'm assuming you are the most qualified person to advice :) ?

>  Could you remind me: is the point of this PR to implement a better solution to hot reloading than the one we currently have or also to improve loading time?

Basically both, since handling the optimization issues wasn't possible with the old hot-reloading implementation.

---

But yes I agree that the PR is quite big, I wouldn't advice to use github to review it, rather something as pycharm/webstorm git diff view:

image

And of course my offer about the call stil stands :) - https://calendly.com/viktor-yunenko/45min

sebgrebe:
Closing as it's been replaced by #935

PR #933, opened 2020-05-31
Allow to run project without docker

victor-yunenko:


sebgrebe:
We'll park this for now as it might create deployment issues.
If we want to use this capacity, we can use the branch

PR #950, opened 2020-07-15
777 export group data from admin panel

### LINKED ISSUE #777: Access/notifications to personal and group profiles data from the admin panel
Implement: 

> I'd like to be able to export data. What I'd like is easy access to - all group data + group organiser names and group organiser email addresses.

sebgrebe:
Gives admins ability to download groups and profiles as csvs. Closes #777 

How to check functionality in development mode:
- Make sure you have at least one profile and one localgroup in your local database
- Get password for admin account by 
  - clicking on "send me password reset email" on login page
  - request for "admin@example.com"
  - go to localhost:8001 to check emails
  - go on password reset link in last email received
- login as admin with new account
- go to localhost:8000/admin
- go to profiles and localgroups tabs respectively and choose "export as csv" from actions

### DIFF
--- a/eahub/localgroups/models.py
+++ b/eahub/localgroups/models.py
@@ -87,6 +87,14 @@ class LocalGroup(models.Model):
         else:
             return "Other"
 
+    def convert_to_row(self, field_names):
+        values = []
+        for field in field_names:
+            if (field == 'local_group_types'):
+                values.append(self.get_local_group_types())
+            else:
+                values.append(getattr(self, field))
+        return values
 
 class Organisership(models.Model):
     local_group = models.ForeignKey(LocalGroup, on_delete=models.CASCADE)
--- a/eahub/localgroups/models.py
+++ b/eahub/localgroups/models.py
@@ -90,11 +90,12 @@ class LocalGroup(models.Model):
     def convert_to_row(self, field_names):
         values = []
         for field in field_names:
-            if (field == 'local_group_types'):
+            if field == "local_group_types":
                 values.append(self.get_local_group_types())
             else:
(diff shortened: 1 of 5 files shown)


eahub/localgroups/models.py
```diff
@@ -87,6 +87,15 @@ def get_local_group_types(self):
         else:
             return "Other"
 
+    def convert_to_row(self, field_names):
+        values = []
+        for field in field_names:
+            if field == "local_group_types":
+                values.append(self.get_local_group_types())
```
LisaJD, line 94: I can't see 'Organisers' column in the exported csv file. Should it be there? I'm wondering if maybe it's because it's a many to many field in the model so it needs a similar function to return values as a list like has been done for local_group_types?

victor-yunenko:
Btw are you sure you don't want to use [django-import-export](https://github.com/django-import-export/django-import-export)? It basically does the same thing, integratable with one mixin and has 97% tests coverage - sorry for dropping it by a bit too late 😅

eahub/localgroups/models.py
sebgrebe: Good catch! I missed it because it's a many-to-many field, which "meta.fields" does not retrieve. 
Fixed.

sebgrebe:
> 
> 
> Btw are you sure you don't want to use [django-import-export](https://github.com/django-import-export/django-import-export)? It basically does the same thing, integratable with one mixin and has 97% tests coverage - sorry for dropping it by a bit too late 😅

Not quite sure how it would handle many-to-many and enum fields, so I'm fine with sticking what we have atm as it's working. Thanks for the tip though

PR #952, opened 2020-07-19
931 email notifications on edits

sebgrebe:
Sends emails to groups@effectivealtruism.org and lean@eahub.org on the following actions:
- Group is created  
- Group is updated
- Group is deleted

The emails look as follows (emails addresses are dummy emails, not real ones):
image
image
image  

**How to QA**:
Create, update and delete a group and check whether 
  1) The actions are successful (group created, updated, deleted)  
  2) Emails received at localhost:8001


cafelow:
Perfect! Thanks Seb!

Issue #880, opened 2020-03-17
Add option to put name of organiser into "Organiser" /group/* without them being registered


cafelow:
There do seem to be people who don't want to join the Hub, but do run groups. Having a name is still useful in those cases. 

This would be something that EA Hub admins would add, or potentially admins of the group would do.

sebgrebe:
Some clarification for developers:  
- The organiser field in localgroups type is linked to profiles class. 
- What we want is that one can also assign a name as organiser, which is not linked to a profile  
- A solution to this might be to introduce a new field in the localgroup type (e.g., "OrganiserFreeText"). This should only be changeable for admins from the admin panel. On the frontend, this should be displayed in addition to the normal organiser field (if not empty), but just as text, not as a link to a profile

PR #954, opened 2020-07-26
Free-text organiser field added to localgroup model

LisaJD:
Closes issue #880

- I've added a new field to the model, and on the group detail page it will show the two fields (if both populated) like below:

organiser

- As Sebastian recommended above, the option to add a non-member organiser is not available on the new group form, only from the admin panel (is this definitely what is wanted? So now non-member organisers can only be added on admin panel, but member organisers can only be added on the new group form):

organiser2
organiser3


sebgrebe:
image
Instead of having 2 separate bullet points here, could we instead just list the non-members with the members, but without link?

eahub/templates/eahub/group.html
```diff
@@ -80,7 +80,7 @@ <h1 class="word-break">{{ group.name }}</h1>
         {% if group.city_or_town %}<li class="list-group-item"><b>City/Town: </b>{{ group.city_or_town }}</li>{% endif %}
         {% if group.country %}<li class="list-group-item"><b>Country: </b>{{ group.country }}</li>{% endif %}
         {% if group.public_organisers %}
-        <li class="list-group-item"><b>Organisers:</b>
+        <li class="list-group-item"><b>Organisers (Member):</b>
```
sebgrebe, line 83: I think it would be better to add the non-members to the Organisers list (just without links), instead of having an extra bullet point

LisaJD:
image
Changes made, now both types of organiser are appearing on one line.

sebgrebe (approved):

PR #959, opened 2020-08-23
900 prompt add more detail

### LINKED ISSUE #900: Change process to Hub sign up form
The issue: Some people sign up and don't add anything to their profile, and it is hard for me to know whether to approve them. Some of these are genuine EAs (which I've found out from finding them on facebook), but I'd rather not a) have to search on FB for them, and b) don't want to miss other real EAs that aren't using their real name on FB. 

I think the main problem is that people are only asked for the bare minimum of info to sign up, and aren't funnelled to fill in their profile. 

1. Could under "Name" we add "please include both your first and last name"

image
 
2. After people submit their form could they be sent to their profile immediately (at the moment they go to the main Hub page). Ideally there should also be some message on their profile saying "Please add some details about you. You need to add some information in order for your profile to be approved".

sebgrebe:
Closes #900


cafelow:
That is perfect! Thanks Seb!

PR #967, opened 2020-09-13
941 group regional level

### LINKED ISSUE #941: Add regional level for groups
1. Please change the group type options to 
City
National/Regional
University

2. I'd ideally like the option for people to share their state. This is only applicable to a couple of countries.
Ideally have a new field "State (if applicable)" - if it is free text then that is okay. 

If it is hard to add another field, we could 
Please change "City or Town" to "City and State (if applicable)"

sebgrebe:
* Renames group type "Country" to "Regional/National"  
* Adds new field "Region" for local groups. Where a group is in the US and this field is populated, we validate first whether the user entered a US state (including territories). User gets error message if not. This also guarantees that US states are saved in the same format (full name).  
* (Unrelated to other changes): adds field for non-member organisers to local groups form. Currently, this field is only accessible from admin panel. This way, normal, non-admin organisers can add people here  

Closes #941


eahub/localgroups/models.py
```diff
@@ -15,27 +15,20 @@ class LocalGroupType(enum.Enum):
     COUNTRY = 2
     UNIVERSITY = 3
 
-    labels = {CITY: "City", COUNTRY: "Country", UNIVERSITY: "University"}
+    labels = {CITY: "City", COUNTRY: "National/Regional", UNIVERSITY: "University"}
```
victor-yunenko, line 18: Ideally would be great to rename the constant as well

It's definitely a really small thing but I'm slightly concerned with those in the long run. Although I have the impression that in the next ~6 month we might even make the types dynamic (similarly to tags) and that wouldn't be problem anymore.

victor-yunenko:
Makes sense to me 👌

victor-yunenko (approved):

PR #993, opened 2020-10-31
991 fix admin group bug

sebgrebe:
* Fixes a bug that throws error when downloading csv for a local group where one of the organisers is a user without profile (rare edge case)  
* Adds a test to handle this edge case and one for the normal case. We have no python unit tests atm, so I thought I'd make a start to slowly build up tests


eahub/tests/test_localgroups_model.py
```diff
@@ -0,0 +1,53 @@
...
+
+        self.assertEqual(f"{name1}, {name2}", organiser_names)
+
+    def test_organisers_names_handles_users_without_profiles(self):
+        local_group = LocalGroup()
+        local_group.save()
+        user_without_profile = User()
+        user_without_profile.save()
+        o = Organisership(user=user_without_profile, local_group=local_group)
+        o.save()
```
victor-yunenko, line 49: in the future I recommend to use a django generator (factory boy, model mommy, etc) and let them handle the appropriate fixtures generation

nice function names btw :)

victor-yunenko (approved):

eahub/tests/test_localgroups_model.py
sebgrebe: Interesting, didn't know about these! Looks like they would have saved me a few lines. Will keep in mind :)

Issue #939, opened 2020-06-23
Add messaging form on user and group profiles


michaltrzesimiech:
Picking up where we left here https://github.com/rtcharity/eahub.org/issues/586

We need a messaging functionality. The scope is to make it possible for a logged in user to click a button "message" visible on other users and groups profiles and use a contact form to send an email i.e. using SendGrid (or whatever is good enough and less or equally as expensive). 

At the same time, we should disable the functionality to reveal users' email addresses publicly. 

There will be no functionality to receive replies through the Hub. Responding will happen through users' personal email accounts.

If economic, we should include some spam detection. If we find out this is complex task, we should move this to a separate ticket.

sebgrebe:
@michaltrzesimiech My understanding is that @cafelow wanted to talk to CEA about this. 
My questions:
1) Were there any decisions made? Is there any input from CEA regarding this feature?
2) What is the cost situation on this? How many emails do we expect to be sent and would this get us over whatever threshold we have for what we are currently paying for with sendgrid?
3) @cafelow said that there were people who would like to contribute to this as devs. Should we get in touch with them/they with us before we start working on this?

cafelow:
1. I talked to CEA's groups team, and also the tech team. I'm afraid this isn't in the tech team's priority right now, and isn't within the scope of the Groups team. To clarify, the groups team scope of helping the EA Hub is just the groups list + resources relating to groups. Since this feature isn't particularly useful for /groups (as they normally have a public email address), this isn't something the groups team will fund. They've suggested applying for a Meta grant if we want funding.

vagarwalla:
**USER PROFILE WIREFRAME**
I created a wireframe here (ignore the profile redesign for now, will address in another Issue):
https://whimsical.com/5uUECYJe5vvWXZj5M9HDu8

In terms of User workflow:

1. User clicks "Message" button on Contact profile page
2. Message box pop-up appears
2. Optional: Message box autofills name/email 
3. User types in message
4. User clicks "Send"
5. An email is sent to the Contact
6. A confirmation email is sent to the User
7. User exits out of the message box pop-up using "Close" or the exit icon back to the Contact's page. 

Uncertainties

- Do we want to autofill the name/email or just let autofill do this?
- Do we want the message to be a pop-up or a new page? The design would stay basically the same


Does this look good or am I missing any important information?

vagarwalla:
I've created issue #986 to update profile edit feautre so users can toggle messaging on/off.

PR #1005, opened 2020-11-14
Allow messaging users (when authenticated only)

mattg95:
This allows messaging between users and users, and between users and groups in the EA hub (not opening a separate email window). If a local group doesn't have an email address set, it sends the message to the group organisers(s) emails with the subject line '(user) has messaged (localgroupname)'

#939


eahub/base/views.py
```diff
@@ -244,5 +245,38 @@ def form_valid(self, form):
...
+            else:
+                return "you"
+      
+        subject = f"{current_user.profile.name} sent {addressee(self)} a message through the Effective Altruism hub."
+        send_mail(
+            subject,
+            message,
+            current_user.email,
+            self. get_recipient_email()
+        )
```
victor-yunenko, line 273: keywargs can be used here instead of variables, eg:
```
        send_mail(
            subject=f"{current_user.profile.name} sent {addressee}",
            message=form.cleaned_data["your_message"],
            receiver=self.request.user,
            desitinations=[self.get_recipient_email()]
        )
```

eahub/base/views.py
```diff
@@ -244,5 +245,38 @@ def form_valid(self, form):
...
+        messages.success(
+            self.request,
+            "Your message to " + self.recipient().name + " has been sent"
+        )
+        return redirect("/{0}/{1}".format(type, self.recipient().slug))
```
victor-yunenko, line 278: here we need to use the django `reverse` function

eahub/localgroups/views.py
```diff
@@ -94,6 +94,21 @@ def profile(self):
...
 
+class SendGroupMessageView(SendMessageView):
+    def get_recipient(self):
+        return LocalGroup.objects.get(slug=self.kwargs["slug"], is_public=True)
+
+    # if the group doesn't have a set email address, emails are sent to the group organiser(s)
+    def get_recipient_email(self):
+        local_group = LocalGroup.objects.get(slug=self.kwargs["slug"])
+        if local_group.email == "":
+            return local_group.organisers_emails().split(" ")
```
victor-yunenko, line 105: that probably won't work because the list is using ", " to separate emails

victor-yunenko: you can call something as `[user.email for user in local_group.organisers.all()]`

eahub/base/views.py
```diff
@@ -244,5 +245,38 @@ def form_valid(self, form):
         return redirect("/{0}/{1}".format(type, reportee.slug))
 
 
+class SendMessageView(FormView):
+    template_name = "eahub/message.html"
+    form_class = SendMessageForm
+
+    def recipient(self):
+        return self.get_recipient()
```
victor-yunenko, line 253:
```
def get_recipient(self) -> Union[Profile, LocalGroup]:
    pass
```

eahub/templates/eahub/profile.html
```diff
@@ -54,6 +54,11 @@ <h1 class="word-break">{{ profile.name }}</h1>
         <span><a href="{% url 'account_change_password' %}"><i class="fa fa-pencil-alt"></i> Change Password</a></span>
       </div>
       {% endif %}
+      {% if user.id != profile.user_id %}
```
victor-yunenko, line 57: Might be improved with `request.user.id != profile.user_id`

victor-yunenko (changes requested):

eahub/base/views.py
```diff
@@ -244,5 +245,38 @@ def form_valid(self, form):
...
+        current_user = self.request.user
+
+        # adds the group name to the subject line on emails to groups for clarity
+        def addressee(self):
+            if type == "group":
+                return self.recipient().name
+            else:
+                return "you"
+      
+        subject = f"{current_user.profile.name} sent {addressee(self)} a message through the Effective Altruism hub."
```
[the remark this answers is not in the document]
victor-yunenko: I think you might shorten it, since the name that would be usually used is `EA Hub`?

victor-yunenko:
We would also need to enable the captcha on it

eahub/base/forms.py
```diff
@@ -19,3 +19,17 @@ def clean(self):
             ]  # Will raise a error message
         else:
             return cleaned_data["reasons"]
+
+
+class SendMessageForm(forms.Form):
+    your_message = forms.CharField(
+        label='Your message', max_length=10000, widget=forms.Textarea(attrs={"rows": 5, "cols": 20}))
+
```
victor-yunenko, line 27: do we need a field for the sender email?

eahub/base/views.py
mattg95: this bit I don't understand

victor-yunenko:
Thank you, it looks better!

I reviewed just now and proposed another round of small adjustments 👌

victor-yunenko (changes requested):

mattg95:
The base view was getting quite complicated, so I split some of the code and moved it to the profile and group view instances.
I don't know how to activate the Captcha V2, that hasn't been done. I've also noticed a CSS bug after the page has redirected which I'm going to try and fix.

victor-yunenko:
Thank you for the updates!

I'm deploying it to the stage server at the moment to test & review the result 👍

mattg95:
Hi @victor-yunenko could you merge this? If it works I can move onto making the message button switchable on/off

victor-yunenko:
Few points:

Since anybody can message anyone as I understand - we will need to do go with one of the following options:
1. Protect the `message` view with login and profile approval to send a message.
2. Add a captcha + a checkbox that would allow users to disable the non-authenticated (by default). And a single-click link in email that would allow to disable non-authenticated messages.

eahub/profiles/views.py
```diff
@@ -80,6 +83,23 @@ def get_type(self):
         return "profile"
 
 
+class SendProfileMessageView(SendMessageView):
+    def form_valid(self, form):
+        recipient = Profile.objects.get(slug=self.kwargs["slug"])
+        message: dict = form.cleaned_data["your_message"]
+        sender_email: dict = form.cleaned_data["your_email_address"]
```
victor-yunenko, line 90: are you sure it's those are `dict` btw?

in any case thank you for the annotations - I love that stuff 😁

victor-yunenko:
@mattg95, could you help with addition of a @login_required decorator to the send message view? Just to make sure that we leave the bots any chance here 😅 

Few other points that I think we should move into separate tickets:
- allow users to disable messaging on their profiles
- allow anonymous messaging later on

mattg95:
@victor-yunenko, I've removed the 'dict'. From what I read on stack overflow, the @login_required is not possible to use with a class-based view? So I used a login mixin like so 
Screenshot from 2020-12-22 12-17-56

Also, feel free to change it if it needs, I am not going to be offended:)

victor-yunenko:
Alright, thank you!

For the login required - afaik it kind of possible, but with somehow different syntax, eg
```
@method_decorator(login_required, name='dispatch')
class ProtectedView(TemplateView):
```
?

victor-yunenko:
Oh I didn't notice the mixin on your screenshot 😅 

Sure that works just as well 👍

mattg95:
Oh, I didn't see your last message in time and changed it anyway. I'm running into some sort of algolia error when I try to create a profile to test if again, but hopefully that now works

victor-yunenko:
I believe we've merged in the fix for the algolia issue, can you try to pull in the chagnes?

mattg95:
ok, I've pulled changes from master through and tested messaging with some example profiles, its works

mattg95:
@victor-yunenko is this ok to merge now? If so I will work on

- allow users to disable messaging on their profiles

- allow anonymous messaging later on

on new branches

eahub/profiles/views.py
victor-yunenko: @mattg95  :) ?

sebgrebe:
placed profile messaging within feature toggle so that it won't be available on prod before switching on/off is enabled and this has been communicated to users

sebgrebe (approved):

PR #1021, opened 2020-11-29
1007 update navbar

sebgrebe:
* Changes the style of the existing navbar to look as much as possible as the one on the new profiles page: https://eahub-stage.us.aldryn.io/profiles/. 
* It wasn't possible to use the same code as the one on the new profiles page as this is using bootstrap 5, but we're using bootstrap 3 on all other, "old" pages. As it's not the same code, the styles aren't identical, but just similar. To achieve identical styles would have taken a lot more time, and I don't think it's worth the effort.  
* This MR is a hack to make the UX not too inconsistent between new profiles page and the other pages, but we should implement the same navbar on all pages, implying that we should replace bootstrap 3 with 5  
* When testing mobile/desktop, make sure to refresh after switching back and force between mobile and desktop on the inspector.

### DIFF
--- a/eahub/base/static/global/js/navbar.js
+++ b/eahub/base/static/global/js/navbar.js
@@ -1,20 +1,28 @@
 export function initNavbar() {
-  const navbar = new Navbar($('#burger-btn'), $('#navbar'));
+  const navbar = new Navbar($('#navbar-toggle-icon'), $('#navbar'));
   navbar.toggleMenuOnClick();
   navbar.disappearMenuOnMovingCursorAway();
 }
 
 class Navbar {
-  constructor(burgerBtnHtmlElement, navbarHtmlElement) {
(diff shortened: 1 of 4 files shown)


sebgrebe:
image

sebgrebe:
image

sebgrebe:
image

sebgrebe:
image

sebgrebe:
image

victor-yunenko:
Looks good 👌 

Should we maybe merge into `divio` to test it on the stage? I guess we will be deploying them together?

sebgrebe:
@victor-yunenko, I made a few more changes to make sure that profiles, groups and my_profile are always shown next to toggl if one is not on these pages. 

I've deployed to the azure staging, you can check it out there. Looks fine to me

victor-yunenko (approved):

PR #1023, opened 2020-12-03
Extend & reorder users cause areas

mattg95:
When a user goes to pick a cause area in the GUI drop-down, the options aren't ordered as we want them to be. I'm not sure how to change this as I don't think we can change their order in the model enum without messing up the db.


sebgrebe (changes requested):
Fails because of formatting (black)

sebgrebe:
asd

> 
> 
> When a user goes to pick a cause area in the GUI drop-down, the options aren't ordered as we want them to be. I'm not sure how to change this as I don't think we can change their order in the model enum without messing up the db.

You're right, we can't change the values assigned to the enums as this would mess things up.
However, we could reorder when passing them to the template (see function ```edit_profile_cause_areas``` in profiles/views.py). Now, this would be a bit messy at you'd have to hardcode the mapping from old to new order, but if product specified the order they wanted, I'd say it's worth it

sebgrebe:
Looks good otherwise!

mattg95:
@sebgrebe is this ok to be merged now?

sebgrebe:
> 
> 
> @sebgrebe is this ok to be merged now?

Thanks for figuring out how to reorder! I've changed my mind and think it's better to order alphabetically, so that we don't implicitly take a stance on which cause areas are more important. I went ahead and changed the code accordingly, hope that's OK.

Will merge now. Thanks for your work!

sebgrebe (dismissed):

sebgrebe (approved):

Issue #740, opened 2019-06-07
Sometimes, it takes a few times to click a button for it to work 


michaltrzesimiech:
Last valid for the "Edit" button in a group profile.

taymonbeal:
For what it's worth, this has happened to me too.

victor-yunenko:
Sadly it was indeed the case. I was able to reproduce it accidentally and could debug it.

The reason is that our buttons looked as following:
```html
<div class="btn btn-default">
    <a href="{% url 'localgroups_delete' form.instance.slug %}">Delete this group</a>
</div>
```

While it was supposed to be:
```html
<a href="{% url 'localgroups_delete' form.instance.slug %}">
    <div class="btn btn-default">Delete this group</div>
</a>
```

In other words only the inner part of the buttons (the text) were clickable.

victor-yunenko:
deployed and tested 👌

PR #1036, opened 2020-12-06
fix the non-clickable buttons #740

victor-yunenko:


victor-yunenko:
Hi @sebgrebe, if you could quickly check it that would be great, otherwise I think it makes sense to urgently deploy it until EOD

FYI I tested all 3 pages on the stage - https://eahub-stage.us.aldryn.io/

Issue #995, opened 2020-11-02
Missing organizations in the organization list:


vagarwalla:
Add to the Profile organizations in the organization list (these ones seem more obvious):

Future of Humanity Institute
Center on Long-Term Risk
Animal Ethics
Giving What We Can
LessWrong
Center for Applied Ratioanlity
Happier Lives Institute
Innovations for Poverty Action
J-PAL
All-Party Parliamentary Group for Future Generations 
Centre for Enabling EA Learning and Research
The Legal Priorities Project
Sentience Politics
WANBAM

Less clear if we should add them: 
- Do we want a bunch of area-specific giving platforms?
- Do we want CE incubated charities?
- Do we want GiveWell recommended charities?

Basically do we want the list of organisations listed here: https://forum.effectivealtruism.org/posts/f6kg8T2Lp6rDqxWwG/list-of-ea-related-organisations

Or should we do it on a tagging basis later?

@sebgrebe @victor-yunenko

vagarwalla:
Hey guys! Just realised the EA Hub is missing from the organisation list!  ;D
Can we list as: Effective Altruism Hub.

sebgrebe:
> 
> 
> Hey guys! Just realised the EA Hub is missing from the organisation list! ;D
> Can we list as: Effective Altruism Hub.

We're too humble. I guess this makes the Hub becoming an independent entity official? ;)

sebgrebe:
@vagarwalla to answer your question regarding which ones to add: I'd be happy to add the orgs listed in the forum post (you might also have seen that there is a comment explicilty mentioning the Hub, criticising that we don't have those lists).

Generally, this raises the interesting question of how this should be decided. We don't want to play gate keeper, but of course we also shouldn't add orgs willy-nilly. Ideally, we'd have some criteria written up somewhere on how to decide this.

Probably a good topic for our weekly call.

victor-yunenko:
Personally I would just satisfy the majority of users' requests. The organizations are filterable, and there's not much difference in having 100 or 1000 organizations. As long as the end-users are happy, why not?

sebgrebe:
> 
> 
> Personally I would just satisfy the majority of users' requests. The organizations are filterable, and there's not much difference in having 100 or 1000 organizations. As long as the end-users are happy, why not?

[...]

Issue #949, opened 2020-07-13
Update the list of cause areas


michaltrzesimiech:
We want the list to fit that used at EAG:

- Building EA Communities
- Psychology Of Decision-making
- Global Priorities Research
- Mental Health/happiness
- Global Health And Development
- Global Coordination/peace-building
- Climate Change
- Ai Strategy And Policy
- Farmed Animal Welfare
- Other Policy Reforms
- Ai Safety Technical Research
- Other Existential Risks
- Wild Animal Welfare
- Biorisk Strategy And Policy
- Information Security
- Quantitative Trading

And keep 

- Long-term Future
- Rationality

We need to migrate the existing

"Global poverty" to "Global Health And Development"
"Cause prioritisation" to "Global Priorities Research"
"Meta" to "Building EA Communities"

vagarwalla:
Hi, quick question on this:
Some of these are not exactly cause areas eg. quantitative trading. Just wondering if it makes sense to include quant tradinging in cause areas, or list it as a profession instead?

sebgrebe:
@vagarwalla Those would probably better fit into the "Areas of Expertise" category

vagarwalla:
**Heres the final request as per the conversation**


We want the list to fit that used at EAG:

- Building EA Communities
- Psychology Of Decision-making
- Global Priorities Research
- Mental Health/Happiness
- Global Health And Development
- Global Coordination/peace-building
- Climate Change
- Farmed Animal Welfare
- Wild Animal Welfare
- Other Policy Reforms
- AI Strategy And Policy
- AI Safety Technical Research
- Other Existential Risks
- Biorisk Strategy And Policy
- Information Security

We will keep:
- Long-term Future
- Rationality

We need to Migrate:

- "Global poverty" to "Global Health And Development"
- "Cause prioritisation" to "Global Priorities Research"
- "Meta" to "Building EA Communities"

mattg95:
A couple of points not covered above:
- Under the new EAG naming, "mental health" will change to  "mental health/happiness"
- Instead of " Animal welfare and rights" we now have "Farmed animal welfare" and "wild animal welfare". I think we should rename " Animal welfare and rights" >>> "Farmed animal welfare" and create "wild animal welfare" as a new topic.  This will affect users that have ticked 'animal welfare and rights' in the past, and the ones from the legacy hub as well

PR #1037, opened 2020-12-07
Extend & reorder users organizations

mattg95:
There is a discussion of which charities to include here: https://github.com/rtcharity/eahub.org/issues/995. I have added all the ones that Vaidehi has listed, but no more.

### DIFF
--- a/eahub/profiles/models.py
+++ b/eahub/profiles/models.py
@@ -148,11 +148,13 @@ class GivingPledge(enum.Enum):
 class OrganisationalAffiliation(enum.Enum):
 
     EIGHTY_THOUSAND_HOURS = 1
+    ALLFED = 32
     ANIMAL_CHARITY_EVALUATORS = 2
     BERKELEY_EXISTENTIAL_RISK_INITIATIVE = 3
     CENTER_FOR_APPLIED_RATIONALITY = 4
     CENTER_FOR_HUMAN_COMPATIBLE_AI = 5
     CENTRE_FOR_EFFECTIVE_ALTRUISM = 6
+    CENTRE_ON_LONG_TERM_RISK = 34
     CSER = 7
     CHARITY_ENTREPRENEURSHIP = 8
     CHARITY_SCIENCE_HEALTH = 9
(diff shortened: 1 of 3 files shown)


sebgrebe (changes requested):
Thanks for this.
It would be nice if we could order them alphabetical on the drop down, independent of their enum number

image

victor-yunenko:
> It would be nice if we could order them alphabetical on the drop down, independent of their enum number

FYI with django models we can set default order to the `name` field.

victor-yunenko:
FYI I think you can order it by reordering the python enum members, eg instead of
```python
    THE_LIFE_YOU_CAN_SAVE = 30
    WILD_ANIMAL_INITIATIVE = 31
    ALLFED = 32
```
we can write
```python
    ALLFED = 32
    THE_LIFE_YOU_CAN_SAVE = 30
    WILD_ANIMAL_INITIATIVE = 31
```

mattg95:
@victor-yunenko I've tried reordering the python numbers but it doesn't change the order of the drop down, unless I'm missing something. There is a new issue open moving from enums to models: https://github.com/rtcharity/eahub.org/issues/1068. I could take on this issues as I could then order the choices in the model using `ordering=["name"]`. I could also refactor the solution for adding new cause areas which is currently open https://github.com/rtcharity/eahub.org/issues/949 using the same approach.

victor-yunenko:
@mattg95, interesting - did you try to change the order of attributes along with `labels`?

victor-yunenko:
Probably the migration to models is a bigger feature that we should address later on 👌

victor-yunenko:
hi @mattg95! did you have a chance to try to reorder the labels?

mattg95:
> hi @mattg95! did you have a chance to try to reorder the labels?

@victor-yunenko I have tried arranging the attributes and the labels but the only thing that seems to change the order is to change the actual enum number, which obviously is not what we want. Do you know a way of reordering them?

mattg95:
The way I did it when adding cause areas was to make a function that rearrange them like so:
Screenshot from 2020-12-28 20-13-02

sebgrebe:
@mattg95, having discussed this with @victor-yunenko and @vagarwalla, could we add all the orgs [here](https://www.notion.so/ae330bb9deac4186b7528b84939d5448), in alphabetical order? (To see how to do alphabetical ordering, see my change on your PR for causes areas)

mattg95:
@sebgrebe I've added all the orgs from [here](https://www.notion.so/ae330bb9deac4186b7528b84939d5448). One problem I ran into was overrunning  line length character limits (flake8 errors). I've shortened some of the names of organisations so that we can still have full length labels (e.g. see last commit). Is this ok, or will it mess up the db, like if the names are referenced as a key in the db?

sebgrebe (approved):

PR #1038, opened 2020-12-08
Fix profile edit bug

sebgrebe:
* Clears cache after saving to groups or profiles model, to avoid bug that caused changes on profiles not to show
* Adds default_storage_dsn env variable, with reference to test storage


victor-yunenko:
Looks good, thank you 👍 

The only thing I can mention is that in django the convention is to create a `receivers.py` file and load it in through an AppConfig 👌

Issue #989, opened 2020-10-26
Groups import from Zoho with update & creation of the existing & new groups


victor-yunenko:
- [x] Victor to provide a CSV file example
- [x] receive the complete CSV data (Catherine hopes to get it somewhere within 1-2 weeks)
- [x] confirm that the email is related to the group only, not the organizers, if that's not the case handle it appropriately (spot check)  
- [x] don't use the `type` field, use only the new `types`
- [x] hide the new export if it isn't completely aligned with the old `export` django action to avoid the potential confusion
- [x] document the existing urls and check that nothing returns 404 errors (solution we eventually went with prevent changing URLs of existing groups that were updated
- [x] import on the stage
- [x] compare the stage to prod (spot check that groups have been updated according to data)  
- [x] ensure that the backup is in place and easily restorable
- [x] ensure that the groups are connected to the organizers manually (spot check)

specs:
- we would need to hide a bunch of groups using the `is_active` flag

victor-yunenko:
Hopefully I'll be able to provide the CSV examples (with & without IDs) sometime over this weekend 👌

victor-yunenko:
Hi @cafelow, btw could you remind me about the deadline just in case?

victor-yunenko:
Hi @cafelow, I implemented the new import. I didn't deploy it yet to the stage server, but here's an example of how an import confirmation could looks like, where I updated an email of a group and added a new one.

I also invited you to the spreadsheet that contains two sheets:
- [exported groups](https://docs.google.com/spreadsheets/d/1QuKuDKYcifYU0VDAzD2HF2G1XVBPT8CLReVC__f5Zqo/edit#gid=2092585553), that can be used as an example of how an import can look like
- [a truncated sheet from the exported group with one new group (without the `id` field)](https://docs.google.com/spreadsheets/d/1QuKuDKYcifYU0VDAzD2HF2G1XVBPT8CLReVC__f5Zqo/edit#gid=1993894235)

![image](https://user-images.githubusercontent.com/36416236/97812733-0df39b00-1c8c-11eb-915f-bf3302667b5f.png)

I also improved the admin panel - added columns, filters and fixed the search:

![image](https://user-images.githubusercontent.com/36416236/97812704-ca992c80-1c8b-11eb-9113-550c8d73fc2c.png)

victor-yunenko:
*the screenshot of the import confirmation was incorrect and didn't contain the email update

victor-yunenko:
[...]

PR #1043, opened 2020-12-13
989 group data import fix

sebgrebe:
* Fixes issues with group data import to address #989:  
  * Issue that prevented groups from being updated (rather than newly added) caused by utf-8 BOM  
  * Issue with importing group types (required creating own Widget for Enums)  
  * Maps organiser names to profiles      
* Adds functionality for marking selected groups in admin panel public/not_public
* Adds DateRangeFilter for "last_updated" field

### DIFF
--- a/eahub/localgroups/admin.py
+++ b/eahub/localgroups/admin.py
@@ -143,7 +143,7 @@ class LocalGroupResource(ModelResource):
 
 @admin.register(LocalGroup)
 class LocalGroupAdmin(ImportExportMixin, admin.ModelAdmin, ExportCsvMixin):
-    actions = ["export_csv", "make_not_public"]
+    actions = ["export_csv", "make_public", "make_not_public"]
     list_display = [
         "name",
         "local_group_type",
@@ -186,6 +186,11 @@ class LocalGroupAdmin(ImportExportMixin, admin.ModelAdmin, ExportCsvMixin):
         fieldnames.append("organisers_emails")
         return ExportCsvMixin.export_csv(
             self, request, queryset, fieldnames, "localgroups"
+        )
+
+    def make_public(self, request, queryset, **kwargs):
+        queryset.update(
+            is_public=True
         )
 
     def make_not_public(self, request, queryset, **kwargs):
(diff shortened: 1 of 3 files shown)


eahub/localgroups/admin.py
```diff
@@ -45,41 +62,104 @@ class Meta:
...
+    def import_row(
+        self,
+        row,
+        instance_loader,
+        using_transactions=True,
+        dry_run=False,
+        raise_errors=False,
+        **kwargs,
+    ):
+        row = self.remove_chars_from_row_keys(row, "\ufeff")
```
victor-yunenko, line 81: that's interesting, I'm not sure what it does, should we leave a comment?

eahub/localgroups/admin.py
```diff
@@ -45,41 +62,104 @@ class Meta:
...
-        row["local_group_types"] = self.hydrate_local_group_types(row["types"])
-        return super().before_import_row(row, **kwargs)
+        row["local_group_types"] = [
+            type
+            for type in self.hydrate_local_group_types(row["types"])
+            if type is not None
+        ]
+        organisers_users, organisers_non_users = self.hydrate_organisers(row)
+        row["organisers"] = ",".join(map(lambda x: str(x.id), organisers_users))
+        row["organisers_freetext"] = ",".join(organisers_non_users)
```
victor-yunenko, line 94: wasn't the free text field supposed to be used for summaries? if yes perhaps we can leave a `# TODO` here to remove this code for future imports?

victor-yunenko (approved):
Looks like you figured out import-export in the right way 👍 

I think there might be few things that we might want to improve, yet I would ok with merging it in now and creating a ticket for getting back to them in 1-6 months.

eahub/localgroups/admin.py
sebgrebe: no, that's for organisers who don't have a profile.
Summary is other_info field.

eahub/localgroups/admin.py
sebgrebe: has to do with UTF encoding. The csv comes in with this char for the first columnname, so "id" is "\ufeffid" instead. I tried changing the encoding configuration as suggested [here](https://stackoverflow.com/a/53511822), but it didn't work, so I used this approach. I'll rename the method to make it clearer

PR #1050, opened 2020-12-15
Allow filtering users by giving pledge

### LINKED ISSUE #1049: Allow to filter users in admin panel by donation pledge
GWWC would like all public (!) user data for "self-identified GWWC" user data. 

Implement filter on admin panel for pledge.

sebgrebe:
* Adds filter of users by giving pledge to admin panel  
* Adds filter of users by whether there email is visible to admin panel  
* Fixes bug with profiles export such that values where not matched to correct columns  

Triggered by request from GWWC to get a list of all public users who have taken the pledge. Adding filter to email since we don't want to share hidden emails  

closes #1049

### DIFF
--- a/eahub/localgroups/models.py
+++ b/eahub/localgroups/models.py
@@ -121,6 +121,16 @@ class LocalGroup(models.Model):
                 values.append(getattr(self, field))
         return values
 
+    @staticmethod
+    def get_exportable_field_names():
+        fieldnames = [
+            field.name
+            for field in LocalGroup._meta.fields + LocalGroup._meta.many_to_many
+            if field.name != "local_group_type"
+        ]
+        fieldnames.append("organisers_emails")
+        return fieldnames
+
 
 @receiver(post_save, sender=LocalGroup)
 def clear_the_cache(**kwargs):
--- a/README.md
+++ b/README.md
@@ -47,6 +47,12 @@ docker-compose run --rm web pytest
 npm test
 ```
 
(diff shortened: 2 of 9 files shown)


eahub/localgroups/models.py
```diff
@@ -121,6 +121,16 @@ def convert_to_row(self, field_names):
                 values.append(getattr(self, field))
         return values
 
+    @staticmethod
+    def get_exportable_field_names():
+        fieldnames = [
+            field.name
+            for field in LocalGroup._meta.fields + LocalGroup._meta.many_to_many
+            if field.name != "local_group_type"
+        ]
```
victor-yunenko, line 130: looks tricky :)

although I don't have anything against this code really, I'm just not 100% sure re what it does

Usually I recommend to split it on sub items with clarification variables, to make sure that even the most junior people can understand 100% and there's almost zero chance of them misusing it. I guess we can be a bit more relaxed here, but it does add pressure for solid e2e tests 😁

victor-yunenko: as I understand it's for admin export, hence of course it isn't mission critical either 👌

victor-yunenko (dismissed):

eahub/localgroups/models.py
sebgrebe: I've added a unit test to 1) make intention of the code clearer and 2) decrease pressure for e2e tests

sebgrebe: And added an ananlogous test for Profile

Issue #1031, opened 2020-12-06
Configure a testing pipeline for divio, since we've migrated from azure, try github actions


victor-yunenko:

victor-yunenko:
@sebgrebe, I just tested it - we can also trigger the tests manually now 👍 

![image](https://user-images.githubusercontent.com/36416236/102716577-6a1a8900-42e5-11eb-88b1-497cda4e4b03.png)

PR #1053, opened 2020-12-16
Integrate github actions #1031

victor-yunenko:


sebgrebe:
Bit hard to review just by looking at the code, can I see this in action somewhere? Or we can only use the github actions once this is merged?

sebgrebe (changes requested):
Bit hard to review just by looking at the code, can I see this in action somewhere? Or we can only use the github actions once this is merged?

victor-yunenko:
Sure, they even were running on PRs before, but then I created a new branch (without changes) and they stopped 🤔 

You can see the runs history here - https://github.com/rtcharity/eahub.org/actions

The logs - https://github.com/rtcharity/eahub.org/runs/1559732966?check_suite_focus=true

sebgrebe (approved):
@victor-yunenko demoed and explained how this works to me, looks good.

Issue #1047, opened 2020-12-15
Users are seeing errors when they try to update their cause_areas or careers information - so far 6 users affected


victor-yunenko:
The sentry errors:
- https://sentry.io/organizations/eahub/issues/2093611928/?project=5545943&query=is%3Aunresolved
- https://sentry.io/organizations/eahub/issues/2089133317/?project=5545943&query=is%3Aunresolved
- https://sentry.io/organizations/eahub/issues/2078270888/?project=5545943&query=is%3Aunresolved

The affected email addresses are stored in the error logs.

@vagarwalla, @sebgrebe, I would say this is fairly urgent. If we don't have the time for a fix I would recommend to reach out to the users directly and tell them that we're working on the fix and will keep them posted.

victor-yunenko:
I would be happy to help, but probably no sooner than on Saturday :(

victor-yunenko:
I can confirm, there's chance no one can edit anything at the moment.

victor-yunenko:
Now we need to inform the users, ie:

> Greetings,
> 
> I've received an alert about the server failure that interrupted your profile update. I apologize for the inconvenience and would like to inform you that the problem has been resolved.
> 
> You can now finish the data update - https://eahub.org/profile/
> 
> Thank you for keeping your profile up to date!
> 
> Kind regards,
> Victor

victor-yunenko:
The email has been sent.

PR #1060, opened 2020-12-19
Add a separate algolia index for testing, fix cause_areas or careers update bug, optimize collectstatic and docker

victor-yunenko:
related to #1047

### DIFF
--- a/eahub/config/build_settings.py
+++ b/eahub/config/build_settings.py
@@ -44,5 +44,7 @@ STATICFILES_STORAGE = "whitenoise.storage.CompressedManifestStaticFilesStorage"
 
 ALGOLIA = {
     "APPLICATION_ID": env.str("ALGOLIA_APPLICATION_ID", default="PFD0UVG9YB"),
-    "API_KEY": env.str("ALGOLIA_API_KEY", default=""),
+    "API_KEY": env.str("ALGOLIA_API_KEY", default="83db39f91b7f6a60a901027d0c155bd1"),
+    "API_KEY_READ_ONLY": env.str("API_KEY_READ_ONLY", default="1e1434756fb69ac8fcddd9fcc0ffcdc4"),
+    "INDEX_NAME_PROFILES": env.str("ALGOLIA_INDEX_NAME_PROFILES", default="profiles_stage"),
 }
--- a/eahub/config/build_settings.py
+++ b/eahub/config/build_settings.py
@@ -44,7 +44,7 @@ STATICFILES_STORAGE = "whitenoise.storage.CompressedManifestStaticFilesStorage"
 
 ALGOLIA = {
     "APPLICATION_ID": env.str("ALGOLIA_APPLICATION_ID", default="PFD0UVG9YB"),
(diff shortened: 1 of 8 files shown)


eahub/profiles/models.py
```diff
@@ -530,9 +531,23 @@ def convert_to_row(self, field_names):
         return values
 
     def _format_enum_array_for_searching(
-        self, array: List[enum.Enum], enum_cls: enum.Enum
+        self,
+        enum_values_list: List[Union[enum.Enum, str, int]],
```
sebgrebe, line 535: why do we need this catch all of enums, strings and ints here? All usages of the function seem to be passing enums, not str or int

sebgrebe:
Tested locally, fixes the indexing issue

sebgrebe (approved):
Tested locally, fixes the indexing issue in #1047

eahub/profiles/models.py
victor-yunenko: good point, it's supposed to, but in debugger I noticed that sometimes I a enum value as an int `1` and sometimes as a string `'1'` ¯\_(ツ)_/¯

I think we're thinking about migrating to Model-based FKs, right?

sebgrebe: Super weird. Was this the cause of the bug then? In any case, I'm fine with leaving it like this. 
I'm not aware that we had any discussion around FKs, maybe I missed something?

eahub/config/build_settings.py
```diff
@@ -44,5 +44,7 @@
 
 ALGOLIA = {
     "APPLICATION_ID": env.str("ALGOLIA_APPLICATION_ID", default="PFD0UVG9YB"),
-    "API_KEY": env.str("ALGOLIA_API_KEY", default=""),
+    "API_KEY": env.str("ALGOLIA_API_KEY", default="d1c9139b2271e35f44a29b12dddb4b06"),
```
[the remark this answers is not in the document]
sebgrebe: Discussed this on slack, created ticket: https://github.com/rtcharity/eahub.org/issues/1063

eahub/profiles/models.py
```diff
@@ -530,9 +531,23 @@ def convert_to_row(self, field_names):
...
+        enum_cls: enum.Enum,
     ) -> List[str]:
-        return [item for item in map(enum_cls.label, array)]
+        enum_labels: List[str] = []
```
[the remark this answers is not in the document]
victor-yunenko: hehe, yes I respect their baby-steps with it 😁

eahub/profiles/models.py
victor-yunenko: The cause might be the incorrect enum implementation - I guess the enum package that we're using is outdated since python3.7 introduced the official enum specification, which is significantly different, but in a better way.

> I'm not aware that we had any discussion around FKs, maybe I missed something?

I'm pretty sure we discussed it with you as well on gdrive or github - in short in django the best practice for this use case are FKs or M2M models with enabled native `list_filter` 👍

victor-yunenko: ie similar to how django handles groups & permissions:

image

PR #1062, opened 2020-12-19
Add instructions on how to debug python in docker container

sebgrebe:
@victor-yunenko, feel free to add something on how to do it with PyCharm


victor-yunenko (approved):

victor-yunenko:
FYI I would be happy to assist with the non docker instructions later on 👌

PR #1073, opened 2020-12-20
github actions: fix the directory name, add manual triggers #1031

victor-yunenko:
related to #1031


victor-yunenko:
@sebgrebe, I think there's a bug that prevents github from triggering the action. On the forums devs don't know why it's happening.

I even restored the filesystem state 1:1 to the state where the triggers were working.

I was able to fix that issue by creating the workflow through https://github.com/rtcharity/eahub.org/actions/new - which I haven't done before, but now that it seems to be working.

victor-yunenko:
Nevermind - sorry it was my fault, somehow I renamed the parent directory <_<

PR #1074, opened 2020-12-20
disable algolia when running unit tests

sebgrebe:
- Unit tests have been failing because they made a call to Algolia with incorrect API key.  
- Fix is to disable calling Algolia when running unit tests, shouldn't call external dependency as part of unit tests anyway  
- Some formatting
- Adding ```is_organiser``` to ```Profile``` model, will be used in later PR


eahub/profiles/index.py
```diff
@@ -1,39 +1,43 @@
-from django.conf import settings
 from algoliasearch_django import AlgoliaIndex
 from algoliasearch_django.decorators import register
+from django.conf import settings
 
 from eahub.profiles.models import Profile
 
+if not settings.IS_ENABLE_ALGOLIA:
```
victor-yunenko, line 7: looks like the `not` might be not intended?

eahub/config/settings.py
```diff
@@ -193,11 +193,16 @@ class DjangoEnv(Enum):
 ACCOUNT_USERNAME_REQUIRED = False
 
 # search
+IS_ENABLE_ALGOLIA = True
```
victor-yunenko, line 196: I would recommend `env.str("IS_ENABLE_ALGOLIA", default=True)` and then an addition to `.env` as `IS_ENABLE_ALGOLIA=false`, since sounds as we can benefit from running the tests with agolia through github actions? and I personally would run them locally as well

eahub/profiles/index.py
sebgrebe: Doh, got this mixed up, thanks!

eahub/config/settings.py
sebgrebe: Implemented.

victor-yunenko (approved):

eahub/config/settings.py
```diff
@@ -193,11 +193,16 @@ class DjangoEnv(Enum):
 ACCOUNT_USERNAME_REQUIRED = False
 
 # search
+IS_ENABLE_ALGOLIA = env.str("IS_ENABLE_ALGOLIA", default=True)
```
victor-yunenko, line 196: I think it needs to be added to build_settings.py at well 😅

victor-yunenko (changes requested):

eahub/config/settings.py
sebgrebe: done

victor-yunenko (approved):
👍

victor-yunenko:
I'll deploy it to the staging server

PR #1075, opened 2020-12-23
Cleanup and optimize flake8, black and isort configuration by moving it to pyproject.toml

victor-yunenko:


victor-yunenko:
FYI the only change is the ignoring of `node_modules`, otherwise it's simply about refactoring

sebgrebe (approved):

PR #1084, opened 2020-12-29
Add feature flags

sebgrebe:
* Adds the django package [feature-flags](https://github.com/cfpb/django-flags)   
* Allows for toggling new features, for example, showing them locally and on stage, but not in prod (a [pattern proposed by Martin Fowler](https://martinfowler.com/articles/feature-toggles.html)

### DIFF
--- a/eahub/templates/eahub/profile.html
+++ b/eahub/templates/eahub/profile.html
@@ -80,6 +80,13 @@
 </div>
 {% endif %}
 
+{% load feature_flags %}
+{% flag_enabled 'MESSAGING_FLAG' as messaging_flag %}
+{% if messaging_flag %}
+<div>
+  This is the new, currently hidden messaging feature
+</div>
+{% endif %}
 
 <div class="container profile-info profile-cards">
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,144 +4,371 @@
 #
 #    pip-compile requirements.in
 #
-airtable-python-wrapper==0.15.1  # via -r requirements.in
-aldryn-addons==2.0.1      # via aldryn-django
-aldryn-client==2.3.5      # via aldryn-django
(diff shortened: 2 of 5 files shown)


eahub/templates/eahub/profile.html
```diff
@@ -80,6 +80,13 @@ <h1 class="word-break">{{ profile.name }}</h1>
 </div>
 {% endif %}
 
+{% load feature_flags %}
+{% flag_enabled 'MESSAGING_FLAG' as messaging_flag %}
+{% if messaging_flag %}
+<div>
```
sebgrebe, line 86: Temporary example, to be replaced by the actual new messaging feature

requirements.txt
```diff
@@ -49,3 +49,4 @@ https://control.divio.com/api/v1/apps/serve/aldryn-django/2.2.16.1/b460938a-f357
 django-import-export
 us==2.0.2
 django-admin-rangefilter==0.6.3
+django-flags==5.0.1 # for feature flags
```
victor-yunenko, line 52: @sebgrebe, oh btw the flow for the requirements changes:
- add the requirement to requirements.in
- run `docker-compose run --rm web bash -c 'pip-compile requirements.in > requirements.txt'`

I thought about adding it to the readme but then realized it would be better to move to poetry and store everything in pyproject.toml in any case

victor-yunenko (changes requested):

victor-yunenko:
I guess we also need to merge it into the messaging branch since otherwise we couldn't deploy it?

sebgrebe:
> 
> 
> I guess we also need to merge it into the messaging branch since otherwise we couldn't deploy it?

@victor-yunenko My idea was to merge this first and then rebase messaging branch from master. Either way is fine with me though

requirements.txt
sebgrebe: @victor-yunenko, where is requirements.in? I can't see it among the files in the project

victor-yunenko: Ah right, it's another branch. I would recommend merging it in since I wouldn't want to deploy or test it without frozen requirements.

victor-yunenko (approved):
all looks good, thank you 👍

PR #1085, opened 2020-12-29
982 logging profile activity

### LINKED ISSUE #982: Add logging of profile updates, that would allow to analyze the logs in order to assess the profiles activity
## Desired state  
Admins can see which actions users have performed at which time. There are two reasons we'd like this feature:  
* To have more data on our user behaviour that we can analyse to improve the page 
* Potentially use this data to indicate to other users how up-to-date a profile is  

## Acceptance criteria  
* Any time a user updates the profile, e.g., changes the summary or adds an organisation they are associated with, save this in a (to be newly created) model  
* Admins can access this information from the admin panel (eahub.org/admin)  
* Information should be filterable by user  
* Information should be orderable by action date  
* Admins can download a csv with that information from the admin panel (similar to groups or profiles)  
* Fields required in new model:  
  * Date of action  
  * User name/profile    
  * Field changes (e.g., summary, city etc)   

## Context/tips  
* Probably best to use post_save functionality of django model (see @victor-yunenko's [comment](https://github.com/rtcharity/eahub.org/issues/982#issuecomment-723595230)   
* Stakeholder: @vagarwalla (PM, has admin rights) asked to be able to see when a user last updated their profile

sebgrebe:
* Adds model ProfileAnalyticsLog for logging profile updates, with fields:
  * profile: profile that is updated   
  * time: time it has been updated
  * field: field that is been updated
  * value: value it has been updated with
  * old_value: old value that has been overwritten (empty string where no old value present)
  * action: action tupe ("creation" or "update")
  * action_uuid: unique identifier for the action which field has been updated with. For example, a user can change summary and facebook page on the main profile edit page. That update would have 2 logs with the same action_uuid.
* Adds view of logs in admin panel with export functionality   

@vagarwalla, please do UAT on [dev page](https://eahub-dev-815fc48.us.aldryn.io/admin)  
* Admin panel can be found under Profiles/Profile analytics logs  
* To create more logs, edit your own profile

closes #982


eahub/profiles/models.py
```diff
@@ -555,6 +556,45 @@ def clear_the_cache(**kwargs):
     cache.clear()
 
 
+@receiver(pre_save, sender=Profile)
+def on_change(sender, instance, **kwargs):
+    if instance.id is not None:
```
sebgrebe, line 561: this handles updates to existing profiles

eahub/profiles/models.py
```diff
@@ -555,6 +556,45 @@ def clear_the_cache(**kwargs):
...
+                analytics.profile = instance
+                analytics.time = datetime.now()
+                analytics.action = f"Update {field}"
+                analytics.value = value if value is not None else ""
+                analytics.old_value = old_value if old_value is not None else ""
+                analytics.save()
+
+
+@receiver(post_save, sender=Profile)
+def save_new_profile_to_analytics(**kwargs):
```
sebgrebe, line 577: this handles adding new profiles

vagarwalla:
I've changed every field in my profile and it seems to be working good :slightly_smiling_face:
Small requests for analytics reasons, LMK if they're too much trouble:
Aggregated view by unique users updating at least once in a given time frame e.g. in August, 500 unique users updated their profile, 100 created profiles
Can we have the "Value" field renamed to "New Value" so it's more clear that it's New value/old value
Right now the code value (is that what it's called?) is showing for drop-down changes - e.g. change to career interest areas. Can we replace with the English words? (label value?)

victor-yunenko:
The aggregated amount of updates per month per user - https://docs.google.com/spreadsheets/d/1LsAIUL7TBWulN1RiX8QpeX7SODwIwoL__XxFZZ8phAk/edit?usp=sharing

To do that firstly we need to add `=MONTH(DATE_COLUMN)` column, and then configure the pivot table.

eahub/tests/test_profile_models.py
```diff
@@ -57,3 +52,95 @@ def test_get_exportable_field_names(self):
...
+        )
+        analytics_logs_user_id = ProfileAnalyticsLog.objects.filter(
+            profile=profile, field="user_id"
+        )
+        analytics_logs_email_visible = ProfileAnalyticsLog.objects.filter(
+            profile=profile, field="email_visible"
+        )
+
+        self.assertEqual("User1", analytics_logs_name[0].value)
+        self.assertEqual("False", analytics_logs_is_approved[0].value)
```
victor-yunenko, line 87: FYI you can do `ProfileAnalyticsLog.objects.filter(profile=profile, field="is_approved").first() :)

eahub/tests/test_profile_models.py
```diff
@@ -57,3 +52,95 @@ def test_get_exportable_field_names(self):
...
+        )
+        self.assertTrue(
+            all(x.time == analytics_logs_update[0].time for x in analytics_logs_update)
+        )
+
+
+def create_profile(email, username):
+    user = User()
+    user.email = email
+    user.save()
```
victor-yunenko, line 139: and another nice shortcut
```python
user = User.objects.create(email=email)
```

sebgrebe:
> * Can we have the "Value" field renamed to "New Value" so it's more clear that it's New value/old value
> 
> * Right now the code value (is that what it's called?) is showing for drop-down changes - e.g. change to career interest areas. Can we replace with the English words? (label value?)

- Added names for enums 
- replaced "value" with "new_value"
- Added changes of group membership to logs, too

sebgrebe:
@victor-yunenko, implemented your comments, please approve if it's otherwise good to go

eahub/profiles/models.py
```diff
@@ -566,9 +566,33 @@ def get_exportable_field_names():
         ]
 
 
-@receiver(post_save, sender=Profile)
-def clear_the_cache(**kwargs):
-    cache.clear()
+class ProfileAnalyticsLog(models.Model):
+    profile = models.ForeignKey(Profile, on_delete=models.CASCADE)
+    time = models.DateTimeField()
```
victor-yunenko, line 571: To reduce the maintenence effort do you we could replace the `store` method at the bottom with `time = models.DateTimeField(default=timezone.now)`? And then call `ProfileAnalyticsLog.objects.create(profile=profile, ...)`?

eahub/profiles/receivers.py
```diff
@@ -0,0 +1,88 @@
...
+    action_uuid = uuid.uuid4()
+    time = datetime.utcnow().replace(tzinfo=pytz.utc)
+    for (field, value) in instance.__dict__.items():
+        if (value and field not in profile_fields_to_ignore) or value is False:
+            log = ProfileAnalyticsLog()
+            log.store(
+                instance,
+                field,
+                "Create",
+                "",
```
victor-yunenko, line 37: that's a nice place for the keyword arguments :)

eahub/profiles/receivers.py
```diff
@@ -0,0 +1,88 @@
...
+    if value is None:
+        return ""
+    if type(value) is not list:
+        return value
+    elif len(value) == 0:
+        return ""
+    elif field in profile_fields_enums_map.keys():
+        return [profile_fields_enums_map[field].get(int(x)).label for x in value]
+    else:
+        raise Exception(f"Value {value} cannot be made printable")
```
victor-yunenko, line 73: I would recommend logger.error() here that will alert sentry, and return an empty string instead. As I understand otherwise a runtime failure would result in a server error for the user?

eahub/profiles/receivers.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+@receiver(post_save, sender=Profile)
+def clear_the_cache(**kwargs):
+    cache.clear()
+    if "created" in kwargs.keys() and kwargs["created"]:
+        save_logs_for_new_profile(kwargs["instance"])
```
victor-yunenko, line 80: perhaps we can create another `@receiver` for the logs? for debugging it can be helpful when a function as `clear_the_cache` only clears the cache :)

eahub/profiles/views.py
```diff
@@ -178,18 +179,37 @@ def edit_profile_community(request):
     if request.method == "POST":
         form = EditProfileCommunityForm(request.POST, instance=request.user.profile)
         if form.is_valid():
+            old_local_groups = [
+                x.name
+                for x in Profile.objects.filter(id=request.user.profile.id)[
+                    0
+                ].local_groups.all()
+            ]
```
victor-yunenko, line 187: Can we do something as `[group.name for group in LocalGroup.objects.filter(membership__profile= request.user.profile)]?`

I would also recommend to do it on the line №208 to follow the variable lifespan guidelines (the variable is be defined as close to its usage as possible to minimize the cognitive load of the programmer)

eahub/profiles/receivers.py
```diff
@@ -0,0 +1,88 @@
...
+    "giving_pledges": GivingPledge,
+    "expertise_areas": ExpertiseArea,
+    "organisational_affiliations": OrganisationalAffiliation,
+}
+
+
+def save_logs_for_new_profile(instance):
+    action_uuid = uuid.uuid4()
+    time = datetime.utcnow().replace(tzinfo=pytz.utc)
+    for (field, value) in instance.__dict__.items():
```
victor-yunenko, line 30: do you think we can use `for field in Profile._meta.get_fields()`? that's the method django is using in their codebase

and then perhaps we can call `getattr(instance, field.name)`

eahub/profiles/receivers.py
```diff
@@ -0,0 +1,88 @@
...
+                convert_value_to_printable(value, field),
+                time=time,
+                action_uuid=action_uuid,
+            )
+
+
+def save_logs_for_profile_update(instance, previous):
+    new_fields = instance.__dict__.items()
+    action_uuid = uuid.uuid4()
+    time = datetime.utcnow().replace(tzinfo=pytz.utc)
```
victor-yunenko, line 47: Do you think django's `timezone.now()` is doing the same as `datetime.utcnow().replace(tzinfo=pytz.utc)`? I haven't seen that later before

victor-yunenko (changes requested):
the architecture makes sense 👍

I found small things here and there that I left comments about 😅

eahub/profiles/models.py
victor-yunenko: oh actually there's `models.DateTimeField(auto_now_add=True)`

eahub/profiles/receivers.py
sebgrebe: changed

eahub/profiles/models.py
sebgrebe: Used (default=timezone.now)

eahub/profiles/receivers.py
sebgrebe: This can't handle our enum fields:

        profile = create_profile("test@email.com", "User1")

        profile.name = "User1New"
        profile.cause_areas = [CauseArea.BUILDING_EA_COMMUNITIES]
       profile.save()
>       TypeError: cannot unpack non-iterable ManyToOneRel object

victor-yunenko (approved):

PR #1095, opened 2021-01-10
profiles: fix the email link on the vuejs list

victor-yunenko:


sebgrebe (dismissed):
Fine for now, but when we toggle on messaging, we want to get rid of this or replace it with a link to the messaging page, don't we?

victor-yunenko:
I would say not necessary, as some people still might want to share their emails, eg group organizers

sebgrebe:
> 
> 
> I would say not necessary, as some people still might want to share their emails, eg group organizers

At a minimum we need to redesign it as we're using the "email" icon for messaging, so we can't use it twice

sebgrebe (approved):

Issue #459, opened 2019-03-19
Implement SPAM detection, eg by rolling back to recaptcha v2 and adding v3 later on


michaltrzesimiech:

taymonbeal:
What's the use case here? Is it prevention of automated spam, or something else?

michaltrzesimiech:
We seem to have cases of bogus accounts and I'm afraid that these will be back after we delete them. Blocking by IP or whatever makes more sense could also be useful, also for blocking automated spam after we introduce messaging. And also, after we add the "report abuse" button, we should have a flow for excluding users, ideally with some audit trail.

taymonbeal:
I suspect we should use an actual spam detection API like Akismet for this.

taymonbeal:
What of this do we actually need before the freeze?

michaltrzesimiech:
A spam detection API would be great. Based on what we observed so far, we've had something like 5 new spam profiles created since the official launch. We can control this amount manually, so this can be postponed until post-freeze.

michaltrzesimiech:
Prioritizing this for https://github.com/rtcharity/eahub.org/milestone/8 since this issue will affect that.

sebgrebe:
Need to add spam report buttons for admins as discussed here

![Screen Shot 2019-08-23 at 16 49 02](https://user-images.githubusercontent.com/15908208/63605480-f3249180-c5c5-11e9-9eda-096fdb914eed.png)

sebgrebe:
What should happen on clicking this button? Email sent to admins?

taymonbeal:
It should delete the item and send it to Akismet as an example of spam. In order to do the latter, we need to integrate Akismet.

michaltrzesimiech:
We should prioritise this higher since the manual maintenance costs accumulate with each month

vagarwalla:
Some users were blocked because the wrong pattern was blocked (e.g. .info/.iu)

victor-yunenko:
It has been deployed and tested.

We have a new google admin now. I can add more google accounts on request.

PR #1101, opened 2021-01-10
migrate to captcha v2 - fix envs loading #459

victor-yunenko:
#459


victor-yunenko:
@sebgrebe, there was a bug I think that made it testable in dev env but won't work in prod. Could you help with one more review?

PR #1102, opened 2021-01-10
profile: fix the non clickable profile urls links

victor-yunenko:


victor-yunenko:
Hi @sebgrebe, I need to merge it - that's a regression, where after a merge conflict resolution on the messaging branch external parts of the buttons became non-clickable again.

I guess I'll need to force merge it now.

cc @mattg95

victor-yunenko:
Tested on the stage - https://eahub-stage.us.aldryn.io/profiles/?

PR #1110, opened 2021-01-13
1018 add feedback box

LisaJD:
Added so far:

Separate sub app with all functionality in one app, template added to templates/components/
Functionality matching example website + EA colour scheme
Added to eahub.org/profiles
Data saving in django model - message required but email optional
On small screens appears smaller and without close button

Still left to implement:

1. Show 'Feedback has been sent' message (is the best approach to use 'django.contrib.messages'?), and don't reload page

2. Save current page url into model correctly
Currently request.path is saving the /feedback/submit endpoint into the form, not /profiles - not currently sure how to fix this

3. Form validation to not let user send blank form - I'm looking at ReportAbuseView for example of how to do this


eahub/base/models.py
```diff
@@ -1,4 +1,5 @@
 from authtools import models as authtools_models
+from django.db import models
```
LisaJD, line 2: Didn't mean to add this - will remove in next push

sebgrebe:
Design looks great and basic functionality is working, thanks! I'll leave a couple of comments

eahub/feedback/migrations/0001_initial.py
```diff
@@ -0,0 +1,30 @@
...
+                (
+                    "id",
+                    models.AutoField(
+                        auto_created=True,
+                        primary_key=True,
+                        serialize=False,
+                        verbose_name="ID",
+                    ),
+                ),
+                ("email", models.CharField(blank=True, max_length=30)),
```
sebgrebe, line 25: We should add creation time as a field so that we know when the feedback was made

eahub/feedback/admin.py
```diff
@@ -0,0 +1,5 @@
+from django.contrib import admin
+
+from .models import Feedback
+
+admin.site.register(Feedback)
```
sebgrebe, line 5: image

This looks very basic. It would be great if this could look similar to the other admin panels: 
- columns for: email address, comment, and time feeback was given (see comment re time on model below).
- message and emails are searchable
- feedback can be ordered by time 
- Admins should be able to export the feedback into a csv.

You can find an example of how to this in ProfileAnalytics: https://github.com/rtcharity/eahub.org/blob/c06d994e4771502678c186fa206aec5a3fdb0616/eahub/profiles/admin.py#L73

Make sure to use ```ImportExportMixin```, rather than utils.ExportCsvMixin for exporting (we unfortunately have 2 different approaches in the code, where possible we should use the external library ```ImportExportMixin```

sebgrebe:
>Show 'Feedback has been sent' message (is the best approach to use 'django.contrib.messages'?), and don't reload page  
Messages is what we're using elsewhere for this, yes. I would suggest that the success message shows in the feedback modal itself, rather that on the main page

eahub/feedback/models.py
```diff
@@ -0,0 +1,7 @@
+from django.db import models
+
+## added by Lisa
+class Feedback(models.Model):
+    email = models.CharField(blank=True, max_length=30)
```
sebgrebe, line 5: why 30? Emails can have up to 254 chars: https://stackoverflow.com/a/1199238

eahub/feedback/models.py
```diff
@@ -0,0 +1,7 @@
...
+    message = models.CharField(max_length=2000)
+    page_url = models.CharField(blank=True, max_length=50)
```
sebgrebe, line 7: 50 is too low. Max length varies with browser, I think we should go at least with 2048 (yes, we probably will never have urls like that, but we don't have to be concerned about data volume here): https://stackoverflow.com/a/33733386

eahub/feedback/views.py
```diff
@@ -0,0 +1,19 @@
...
+        form = FeedbackForm(request.POST)
+        if form.is_valid():
+            form = form.save(commit=False)
+            form.page_url = url
+            form.save()
+            return HttpResponseRedirect("/profiles")
+        else:
+            form = FeedbackForm()
+            return HttpResponse()
+            # display message saying it wasn't valid
```
sebgrebe, line 19: ?

eahub/templates/components/feedback.html
```diff
@@ -0,0 +1,144 @@
+{% block feedback %}
+
+<style>
```
sebgrebe, line 3: Let's separate css from html, add css to main.css

eahub/templates/eahub/profiles.html
```diff
@@ -6,8 +6,10 @@
   {% include 'components/navbar.html' %}
 {% endblock navbar %}
 
+
```
sebgrebe, line 9: ?

sebgrebe (changes requested):

sebgrebe:
image

On mobile, this takes up quite a bit of space and covers text. This is similar in the [example](https://optimus.foundation/en/projects/13/) we used as a guide. 

Is this OK? Do we want the button to be smaller on mobile, the text margin's wider? @victor-yunenko @vagarwalla

sebgrebe:
> On small screens appears smaller and without close button   

We need a close button though, right? It doesn't seem to close if I click outside of the modal. If you worried about the space it would take up, I think we can use the same strategy as the modal filling the full screen, just like in the [example](https://optimus.foundation/en/projects/13/)

viktor-yunenko (changes requested):

victor-yunenko:
> Is this OK? Do we want the button to be smaller on mobile, the text margin's wider? @victor-yunenko @vagarwalla

I would recommend trying to position it at the bottom, eg as `bottom: 0`

victor-yunenko:
FYI you can use the native es6 functions for the form submission, eg

```javascript
fetch(form.action, {
  method: form.method,
  body: new URLSearchParams(new FormData(form)),
});
```

eahub/templates/components/feedback.html
```diff
@@ -0,0 +1,144 @@
...
+    <div class="modal fade" id="feedbackModal" tabindex="-1">
+      <div class="modal-content">
+          <div class="modal-header">
+            <h5 class="modal-title">Leave Feedback</h5>
+          </div>
+          <div class="modal-body">
+            <p>Do you have any feedback for us?</p>
+            <form class="form-group" method="post" action="{% url 'feedback:submit_feedback' %}">
+              {% csrf_token %}
+              <textarea class="form-control" name="message" cols="37" rows="5"></textarea>
```
victor-yunenko, line 114: @LisaJD FYI you can add the `required` attribute here to enable the native browser validation, same with the <input> below 👍

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required

LisaJD:
Recent commit has added the following features:

1. Admin app
Entries are displayed with creation date, email and message columns. Searchable by message or email, can order by date

2. AJAX form submission
Added AJAX to submit form without refreshing page. Shows a success message on submission and hides form, or an error message if feedback field left blank

3. Moved CSS out of template and into an scss file, although haven't implemented any mixins or scss features yet

4. Mobile version
Re-added the close button. Made "Feedback?" button smaller on mobile, and made modal fill entire screen to match the example website. 
Didn't yet implement moving button to bottom of page because it would involve restyling the modal - (was the idea the modal would then pop out from bottom of screen instead of from the right?)

4. Model
Added creation time as a field, changed email max length to 254 and page url max length to 2048

Not yet implemented: 
Saving page url into the model
Some styling on the success/error messages

eahub/templates/components/feedback.html
LisaJD: I added this in the commit today but it doesn't seem to do anything?

eahub/feedback/migrations/0002_auto_20210117_1547.py
```diff
@@ -0,0 +1,31 @@
+# Generated by Django 2.2.16 on 2021-01-17 15:47
```
sebgrebe, line 1: Ideally the 2 migration files would be one with the fields you've added now in the initinal set up.

eahub/feedback/apps.py
```diff
@@ -0,0 +1,6 @@
+from django import apps
+
+
+class FeedbackConfig(apps.AppConfig):
```
sebgrebe, line 4: I was getting 
>>> Model class eahub.feedback.models.Feedback doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.  

when running locally, so I've added  the feedbacks app to installed apps

eahub/feedback/apps.py
```diff
@@ -0,0 +1,6 @@
...
+    name = "eahub.feedback"
+    verbose_name = "Feedback"
```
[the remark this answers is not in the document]
sebgrebe: I assume Lisa simply followed the patterns for other apps, such as profiles and localgroups. Why do you think it's a problem if she does it for feedback?

eahub/templates/components/feedback.html
```diff
@@ -0,0 +1,79 @@
...
+              <input class="form-control" type="text" name="email">
+              <input id="page_url" type="hidden" name="page_url" value="">
+              
+              <div class="modal-footer">
+                <button  id="closebtn" type="button" class="btn btn-secondary" onClick="togglemodal()">Close</button>
+                <input onClick="ajaxsubmit()" value="Submit" class="btn submitbtn" />
+              </div>
+              </form>
+
+              <div id="successmsg" style="display:none">
```
sebgrebe, line 38: image

I can't seem to get out of this view once feedback has been submitted. We need a close button here

sebgrebe: Oh, I see, by clicking on the label. I don't think that's clear enough to the user. Would still add a close button

sebgrebe: When you're done and click on the label again, it still says thanks for your feedback. Is this on purpose? I'd say we'd rather reset in case someone wants to give more feedback.

@victor-yunenko, what do you think? I know you asked Lisa to prevent a refresh here. I assume you meant a refresh of the whole page, but would still prefer it if users could provide more feedback on clicking the label again?

sebgrebe: Finally. when I refresh the page and click on "feedback" again, I see the same feedback I've given before

sebgrebe (changes requested):

eahub/feedback/admin.py
sebgrebe: Admin panel wasn't showing (got swallowed with the adding of admin_reorder to settings.oy). I've added it.

Looks good!

sebgrebe:
Thanks for the updates, @LisaJD! 
I think something went wrong with the rebasing, I've fixed this and pushed to your branch. Please make sure to pull the remote first before continuing to work on this.

I've resolved all my (and @victor-yunenko) conversations that I think have been addressed. Please check the conversations that aren't resolved yet for what still needs to be done.

Apart from that, there is also the issue of the feedback button covering text on mobile (see my comment [here](https://github.com/rtcharity/eahub.org/pull/1110#issuecomment-761557679)). Probably best to follow Victor's suggestion [here](https://github.com/rtcharity/eahub.org/pull/1110#issuecomment-761766993)

eahub/feedback/apps.py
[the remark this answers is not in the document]
victor-yunenko: in this case I don't think this code does something useful, but it has maintenance cost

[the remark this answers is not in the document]
LisaJD: The Django tutorial seemed to suggest an appconfig is required for registering a subapp? But I'll remove it if it's not needed

vagarwalla:
@sebgrebe @victor-yunenko having the feedback button on the bottom on mobile is fine with me!

eahub/feedback/apps.py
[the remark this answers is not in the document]
victor-yunenko: it's true that it's required when you need eg to customize the SQL db prefix, but those use cases are rather rare

eahub/templates/components/feedback.html
```diff
@@ -0,0 +1,144 @@
...
+    background: rgb(99, 82, 116);
+    border-color: rgb(99, 82, 116);
+  }
+</style>
+  
+  <div id="overlay"></div>
+
+  <div id="feedback">
+  <!-- Button trigger modal -->
+  <button type="button" class="btn feedbackbtn" onClick="togglemodal()">
```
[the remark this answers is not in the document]
victor-yunenko: a correction: in bs5 it should be `data-bs-target`, more details can be found here - https://getbootstrap.com/docs/5.0/components/modal/

FYI `onclick` isn't used those days, it's left for the old browsers sake, because there are scripts that silently override it and drop the togglemodal() call :)

eahub/templates/components/feedback.html
victor-yunenko: I like the idea, the form can have a simple success message then

eahub/templates/components/feedback.html
```diff
@@ -0,0 +1,79 @@
...
+      headers: {
+        'Content-Type': 'application/x-www-form-urlencoded'
+      }
+    }).then(response => {
+      if(response.status == "200") {
+        document.getElementById("successmsg").style.display = "block";
+        document.getElementById("form").style.display = "none";
+        console.log("success");
+      }
+      if(response.status == "418") {
```
victor-yunenko, line 71: oh btw in JS it's a requirement to use `===` instead of `==` unless there's a critical reason for the latter - in short `==` is alive mainly for the sake of old browsers as <ie11, since it doesn't really work correctly

LisaJD:
1. Two migration files merged into 1 file

2. Moved feedback button to bottom of screen in mobile view (note: This is working in all browsers except for Safari, any suggestions would be great. Adding position: -webkit-sticky; didn't fix it)

3. After submitting feedback, Close button and button to let user submit more feedback appears. (I thought this would be better than the form refreshing itself if you close and open the modal again, but see what you think)

Screenshot 2021-02-05 at 13 51 02

4. Form clears itself when user clicks 'submit more feedback', so user is presented with a fresh form

5. Page url now saves into model

6. When I removed the app config, everything broke... So I left it there for now

eahub/templates/components/feedback.html
LisaJD: I changed them to ===, but now the submit button has stopped working. Would I need to change the type of "418" from a string to something else if I use ===?

victor-yunenko: quite probably - the comparison coercion there isn't consistent and it's been a bane of javascript for a while, below you can find few examples of how it works in practices:
```javascript
true == '1' // true
true == 'true' // false
0 == '' // true
undefined == null // true
```

victor-yunenko: I guess `response.status` is a `Number`?

PR #1123, opened 2021-01-22
profiles: add more fields for easier moderation

victor-yunenko:
I also reordered the admin panel a bit, that's what we usually do for every project since by default django admin isn't user-friendly.

@sebgrebe, would you be able to help with merging it in on Saturday? I can help with moderation tomorrow then. With this patch I also noticed users who were approved by mistake several months ago (they have spam URLs in their summaries)

image


sebgrebe (approved):
Code looks good, didn't do UAT though.
Maybe we could deploy to dev so that @vagarwalla can test whether this works for her?

PR #1131, opened 2021-01-24
986 messaging on off

mattg95:
- I was initially working on the wrong branch, hence the merge request

- I realised the name of the recipient in "send a message to {name}", in the message.html file was not coming through, on either  group or profile messaging. This is why I've added a method to the group and profile sendMessageView

- I have hidden the 'send message' button on a profile if they have unticked 'allow messaging'. However, the message screen is still accessible if you type in the url `profile/profileName/message` . I'm not sure what the best way to solve this is. Do we want to blog this screen completely for users who don't want to be messaged, or have a screen that says 'this user doesn't want to be messaged'?


sebgrebe:
> I have hidden the 'send message' button on a profile if they have unticked 'allow messaging'. However, the message screen is still accessible if you type in the url profile/profileName/message . I'm not sure what the best way to solve this is. Do we want to blog this screen completely for users who don't want to be messaged, or have a screen that says 'this user doesn't want to be messaged'?  

@mattg95, 404 seems fine to me here. Most important is that you cannot get on the page (because of bots, as @victor-yunenko pointed out)

mattg95:
@sebgrebe should raise a 404 now for profile/message URL if the user has unticked 'allow' messaging

sebgrebe (changes requested):
I suggest you address my comments here on PR 1091 and we can remove this PR (to make things a bit easier)

sebgrebe:
Replaced by this PR: https://github.com/rtcharity/eahub.org/pull/1131

PR #1133, opened 2021-01-26
1091 messaging ux

mattg95:
@sebgrebe I've added the 'name' field and changed the labels as per the mockup (https://whimsical.com/ea-hub-community-directory-improvements-5uUECYJe5vvWXZj5M9HDu8) . How much do you want me to match the styling for the 'send message' page, if we're not doing a modal? I'm not sure how to change the styling without editing /form.html, which would change everything else that extends that.


eahub/base/forms.py
```diff
@@ -19,3 +19,13 @@ def clean(self):
             ]  # Will raise a error message
         else:
             return cleaned_data["reasons"]
+
+
+class SendMessageForm(forms.Form):
+    your_name = forms.CharField(label="Name", max_length=10000, widget=forms.TextInput)
```
sebgrebe, line 25: Can we prepopulate this with their user name? This makes the experience smoother for the user and also makes clear *whose* name we're asking for here. Alternatively, change "name" to "Your name"

sebgrebe, line 25: Can we prepopulate this with their user name? This makes the experience smoother for the user and also makes clear whose name we're asking for here. Alternatively, change "name" to "Your name"

eahub/base/forms.py
```diff
@@ -19,3 +19,13 @@ def clean(self):
...
+    your_email_address = forms.CharField(
```
sebgrebe, line 26: Same here: can we prepoluate with their email address?

eahub/templates/eahub/group.html
```diff
@@ -23,18 +23,17 @@
...
-          <i class="fa fa-envelope"></i> Message
-        </div>
-      </a>
+      {% if request.user.id is not None %}
+      <div class="btn btn-default purple">
+        <a href="{% url 'message_group' group.slug  %}"><i class="fa fa-pencil-alt"></i> Send Message</a>
+      </div>
+      {% endif %}
+
+      {% if group.facebook_group %}
```
sebgrebe, line 32: Let's remove the facebook group "message" button, we don't really need it anymore

eahub/templates/eahub/edit_messaging_preferences.html
```diff
@@ -0,0 +1,27 @@
+{% extends 'form.html' %}
```
sebgrebe, line 1: Looking at your initial implementation of this with a simple check box on the main profile editing page, I think this is actually enough and we don't need this extra page (sorry for changing my mind here). I would simply add the info that people will not know your email address to the check box.

sebgrebe (changes requested):
All in all this looks great!
I would suggest that you address my comments here and on PR 986 together on this PR and we can get rid of PR 986

eahub/base/forms.py
sebgrebe, line 25: Name and email max length can be much shorter.
Email: 254
Name: 200

sebgrebe (commented):
> 
> 
> @sebgrebe I've added the 'name' field and changed the labels as per the mockup (https://whimsical.com/ea-hub-community-directory-improvements-5uUECYJe5vvWXZj5M9HDu8) . How much do you want me to match the styling for the 'send message' page, if we're not doing a modal? I'm not sure how to change the styling without editing /form.html, which would change everything else that extends that.

It's good as it is, I've just added 2 comments on prepopulating name and email field

eahub/templates/emails/message_profile.html
```diff
@@ -0,0 +1,9 @@
+<p>
```
sebgrebe, line 1: Am I right that we're currently not using this, but instead the txt file? You're probably in the middle of this.

I'd suggest that for now, we make "Report an issue" an mailto: link to admin@eahub.org. We can implement a form at a later point. Issue: https://github.com/rtcharity/eahub.org/issues/1139

eahub/templates/eahub/group.html
sebgrebe: done

eahub/base/forms.py
sebgrebe: done

sebgrebe:
One more thing: we need to make sure that messaging is enabled by default

sebgrebe (dismissed):

sebgrebe (dismissed):

sebgrebe (approved):

eahub/profiles/models.py
```diff
@@ -463,8 +464,10 @@ def is_searchable(self) -> bool:
     def get_absolute_url(self):
         return urls.reverse("profile", args=[self.slug])
 
-    def get_email_searchable(self) -> Optional[str]:
-        return self.user.email if self.email_visible else None
+    def messaging_url_if_allowed(self) -> str:
+        if self.allow_messaging:
+            return f'/profile/{self.slug}/message'
```
victor-yunenko, line 469: @mattg95, we will need to replace it with a `reverse`, could you help here?

sadly hardcoded URLs stop silently working in the future

eahub/templates/eahub/profile.html
```diff
@@ -37,31 +37,22 @@
...
-            </div>
-        </a>
-        <a href="{% url 'account_change_password' %}">
-            <div class="btn btn-default purple">
-                <span><i class="fa fa-pencil-alt"></i> Change Password</span>
-            </div>
-        </a>
+      <div class="btn btn-default purple">
+        <span><a href="{% url 'edit_profile' %}"><i class="fa fa-pencil-alt"></i> Edit</a></span>
+      </div>
```
victor-yunenko, line 42: @mattg95, @sebgrebe, we must be highly careful here - this is a button 20% of which isn't clickable. I fixed them thought the site one by one before, but now we have to prevent those occurrences in the future

eahub/templates/emails/message_profile.html
```diff
@@ -0,0 +1,12 @@
+<p>Dear {{ recipient }},</p>
+<p>{{ sender_name }} has sent you a message:</p>
+<p>{{ message }}</p>
+<p>Thanks,</p>
+<p>EA Hub</p>
+<p>
+  <a href="feedback.eahub.org">Feedback</a>
+  <span> | </span>
+  <a href="eahub.org/profile/">Messaging Preferences</a>
```
victor-yunenko, line 9: we need a `{% url %}` tag here too

victor-yunenko: also hardcoding eahub.org isn't safe for various reasons, it has to be a variable

victor-yunenko (changes requested):

eahub/templates/eahub/profile.html
sebgrebe: Thanks for catching this! This wasn't intentional, the branch probably hadn't been rebased for a while.

eahub/templates/emails/message_profile.html
sebgrebe: You're ofc right. Replaced all of the hard coded values with variables

eahub/base/utils.py
```diff
@@ -13,6 +14,10 @@ def user_display(user):
     return profile.name
 
 
+def get_feedback_url(request):
+    return f"https://feedback.{get_current_site(request).domain}"
```
victor-yunenko, line 18: thank you for helping here!

AFAIK sadly get_current_site + request isn't as reliable as I wish it were, I don't remember the details, but I believe the request might contain not enough data in few cases and throw errors. Not 100% sure about it thought, hence just warning.

In the end I had to rely on divio.com env variable as `DOMAIN` or add it manually on non-divio/heroku projects

eahub/localgroups/views.py
```diff
@@ -107,12 +109,17 @@ def form_valid(self, form):
         recipient = self.profile()
         sender_name = form.cleaned_data["your_name"]
         sender_email_address = form.cleaned_data["your_email_address"]
+        env = environ.Env()
+        admin_email = list(env.dict("ADMINS").values())[0]
+        feedback_url = get_feedback_url(self.request)
```
victor-yunenko, line 114: could settings.ADMINS work? I'm asking since in test cases overrides are often utilized, eg [on a parent class](https://github.com/rtcharity/eahub.org/blob/master/eahub/tests/e2e/cases.py), and as I understand the override wouldn't work with the direct envs reading

victor-yunenko (changes requested):

PR #1134, opened 2021-01-26
profiles: allow to drop profiles & related users + restore base.User model display

viktor-yunenko:


eahub/config/settings.py
```diff
@@ -317,7 +317,7 @@ class DjangoEnv(Enum):
         "app": "sites",
         "label": "Website administration",
         "models": [
-            {"model": "auth.User", "label": "User accounts"},
+            {"model": "base.User", "label": "User accounts"},
```
sebgrebe, line 320: What's the value of having user accounts and user account email addresses? Wouldn't the first suffice?

eahub/profiles/admin.py
```diff
@@ -30,7 +34,7 @@ def queryset(self, request, queryset):
 
 @admin.register(Profile)
 class ProfileAdmin(admin.ModelAdmin, utils.ExportCsvMixin):
-    actions = ["export_csv", "approve_profiles"]
+    actions = ["export_csv", "approve_profiles", "delete_profiles_and_users"]
```
sebgrebe, line 37: deletion doesn't seem to work. I've tried to delete my test user, but when I come back to the profile admin panel it's still there:
image

sebgrebe (changes requested):

eahub/config/settings.py
victor-yunenko: I don't see value in it, but the multiple email per user feature already exists and it can't be disabled with a simple patch now

eahub/profiles/admin.py
victor-yunenko: ok thanks, I'll check

sebgrebe (approved):

PR #1136, opened 2021-01-29
add a command for cleaning private data from db

victor-yunenko:


eahub/base/management/commands/remove_sensitive_data_from_db.py
```diff
@@ -0,0 +1,32 @@
...
+from django.core.management import base
+from faker import Faker
+
+from eahub.base.models import User
+from eahub.profiles.models import Profile
+from eahub.profiles.models import ProfileAnalyticsLog
+
+
+class Command(base.BaseCommand):
+    def handle(self, *args, **options):
```
sebgrebe, line 11: How do we make sure that this command isn't accidentally run on a production db?

sebgrebe (changes requested):

eahub/base/management/commands/remove_sensitive_data_from_db.py
victor-yunenko: @sebgrebe, what scenario are you assuming exactly?

sebgrebe: Someone could ssh into the prod container and run it. But I see where you're coming from, they could just as well do some other mayhem there with sql commands, for example, the function doesn't really increase the likelihood of that happening (either someone managed to get into the container not knowing what they're doing/with bad intentions or not). So you're right, it's probably fine to leave it

sebgrebe (approved):

eahub/base/management/commands/remove_sensitive_data_from_db.py
victor-yunenko: ah I guess, I agree we should handle it on the control panel side then

FYI - I'll be looking more into heroku very soon, and it should have granular permissions for as well 👍

PR #1156, opened 2021-02-14
Document data protection and security processes

viktor-yunenko:


docs/data-protection-policy.md
```diff
@@ -0,0 +1,26 @@
+# Data Protection Policy
```
sebgrebe, line 1: Could you update this with the google doc we've created, @victor-yunenko?

sebgrebe (changes requested):

docs/data-protection-policy.md
sebgrebe: @victor-yunenko, I've updated this with our google doc. Could you double check whether this is OK as it is now?

Issue #1003, opened 2020-11-11
Allow users to select custom countries, cause areas, organizations and etc, using a stackoverflow-like tag selector


victor-yunenko:
This is certainly a substantial task that should be considered carefully, since it would drain a significant amount of effort from other potentially more impactful improvements.

![image](https://user-images.githubusercontent.com/36416236/98847705-544db480-2459-11eb-82df-e7d9b6246d24.png)

The technical specification is relatively straightforward though:
- django `GenericForeignKey` Tag model with `name` and `description` fields
- django rest api + vuejs frontend with algolia index search
- a notification for the admins that would allow them to drop or merge the tags created by users (duplicates or simply undesirable tags)

victor-yunenko:
There's also the go-to package for django tags - django-taggit, but its interface is primary suitable for admin panel as far as I can see. Yet we might consider it as a deadly cheap alternative.

victor-yunenko:
For frontend we could use eg http://www.vue-tags-input.com/#/

victor-yunenko:
My guesstimate would be 20-50h.

vagarwalla:
- Thanks for the time estimate! I think that this task is probably worth it but we can delay to Q1/Q2 of 2021.
- Recommend not having "custom countries" since we have a country tag already
- Suggestions for the kinds of tags people should add in grey text. E.g. _Use these tags to indicate important organisational, identity, or other affiliations such as your university, school, uncommon cause areas etc._

victor-yunenko:
I agree about dropping `custom countries` - I think the only use case it had was the ability to select several countries at once, which should be done with a multi-select in any case 👌 

Also I recommend to treat that ticket as the first milestone for any field, and discuss what specific fields should be based on it in different issues.

This implementation ideally would be a django package not related to eahub, since we're hardly the only project who could use it.

PR #1167, opened 2021-02-18
Add stackoverflow-like tags #1003

### LINKED ISSUE #1169: Docker-Compose 3.0 "Extends" functionality removed: "docker-compose run" won't work until rewritten without using extends
Current version of docker-compose.yml in use makes load-bearing use of extends. Support for extends was removed as of docker-compose version 3, our compose file is version 3.4. 

I believe this is why docker-compose run... is failing.

See "Removed" from Version 3: https://github.com/docker/docker.github.io/blob/master/compose/compose-file/compose-versioning.md#version-3

Unfortunately, the reference specifications for compose to not make a note that extends was deprecated as of 3.0 (https://github.com/docker/docker.github.io/blob/master/compose/compose-file/compose-versioning.md).

### LINKED ISSUE #1182: Offering and Looking for fields aren't highlighted in the search results
add `ais-snippet` tag

### LINKED ISSUE #1185: Join profile edit pages into one
### Current state  
We have 4 different profile edit pages
- General  
- Community
- Career
- Causes Area

This is suboptimal UX because it means users have to click a lot to change things  

### Desired state
All edit pages on 1 page: /profile/edit so that users have only 1 place to go to  

### Acceptance criteria
- [ ] All profile edit fields on 1 page: /profile/edit
- [ ] Visually Separated into subsection corresponding to the previous one (General, Community, Career, cause area)
- [ ] All other edit pages and URLs removed (profile/edit_profile_career, etc)  
- [ ] Link in verification email should redirect to /profile/edit to nudge people to provide more info (in lieu of depracated "first_visit" url)
- [ ] Remove "first_visit" URL and related message

### LINKED ISSUE #1190: When selecting organisers when editing group, unapproved profiles are shown
image

We should filter unapproved users from this list.

viktor-yunenko:
- related issue #1003 
- dev server - https://eahub-tags-d01de1e.us.aldryn.io/

Before release:  
- [ ] provide an admin URL with the names that had more than 3 names
- [x] test empty tags creation
- [x] add cause_areas_other and `other` to the search
- [x] add fields for cause areas expertise and affiliations
- [x] move tags to the bottom of profile edit list
- [x] check why cause areas contain expertise
- [x] show "topics I speak about" as a tag field
- [x] make a decision on whether to allow creating cause area tags @vagarwalla  

Testing:
- [x] check whether local groups need bootstrap-multiselect
- [x] forms error messages render
- [x] tags & local groups profile logging
- [x] multi browser testing

Optional or future tickets:
- [ ] Other field: toggle on profile edit page for users who have it filled out and those who haven't  
- [x] Add 2-3 sections to profile edit page to make it less cluttered. [Ideas for sections](https://docs.google.com/spreadsheets/d/1Azn3CJ5HDk1PmO3o7olbvI9t8s0NqLvqkqZ-g-rKjmY/edit?usp=sharing_eip&ts=605725e1)  

Related tickets:
- closes #1169
- closes #1185
- closes #1190
- closes #1182

Release aim: Friday, 2nd of April 2021


eahub/base/static/components/mailchimp/mailchimp.js
```diff
@@ -1,17 +0,0 @@
-(function ($) {
```
sebgrebe, line 1: What was the mailchimp code doing and why are we removing it?

eahub/base/static/components/profile-edit-image.js
```diff
@@ -1,47 +0,0 @@
-$(document).ready(() => {
```
sebgrebe, line 1: I assume it's tricky to separate the multiselect stuff out from other profile edit actions? If not, it would be nice to move the non-tag related JS (i.e., non multiselect stuff) at a later stage to reduce the scope of this MR.

If that's too difficult, we need a full QA list of all the things we used to be able to do on editing profile to make sure we're not losing any functionality when migrating from vanilla to vue JS on this.

eahub/config/settings.py
```diff
@@ -50,20 +50,21 @@ class DjangoEnv(Enum):
...
+    "eahub.profiles.apps.ProfilesConfig",
 ]
 
 MIDDLEWARE = [
     "django.middleware.security.SecurityMiddleware",
     "whitenoise.middleware.WhiteNoiseMiddleware",
     "django.middleware.cache.UpdateCacheMiddleware",
     "django_referrer_policy.middleware.ReferrerPolicyMiddleware",
-    "django_feature_policy.FeaturePolicyMiddleware",
+    "django_feature_policy.PermissionsPolicyMiddleware",
```
sebgrebe, line 73: Why this change?

eahub/config/settings.py
```diff
@@ -276,7 +278,16 @@ class DjangoEnv(Enum):
...
 
-SETTINGS_EXPORT = ["WEBPACK_DEV_URL", "DEBUG", "DJANGO_ENV", "ALGOLIA"]
+DOMAIN = env.str("DOMAIN", "localhost:8000")
+
+SETTINGS_EXPORT = [
+    "WEBPACK_DEV_URL",
+    "DEBUG",
+    "DJANGO_ENV",
+    "ALGOLIA",
+    "RECAPTCHA_PUBLIC_KEY",
```
sebgrebe: Why are we adding RECAPTCHA and DOMAIN here?

eahub/profiles/api/serializers.py
```diff
@@ -0,0 +1,130 @@
...
+        fields = [
+            "pk",
+            "name",
+            "types",
+            "count",
+        ]
+
+
+class ProfileSerializer(serializers.ModelSerializer):
+    tags_generic = TagSerializer(many=True, required=False)
```
sebgrebe: Probably to be discussed with product, but what's the purpose of the generic tag?

eahub/profiles/api/views.py
```diff
@@ -0,0 +1,32 @@
...
+    GenericViewSet,
+):
+    queryset = Profile.objects.filter(user__is_active=True)
+    serializer_class = ProfileSerializer
+
+
+@api_view(["POST"])
+@permission_classes([IsAuthenticated])
+def create_tag_view(request: Response):
+    # todo handle duplicates
```
sebgrebe: Is this about the "merge" actions we talked? So do you want duplicates to be handled code or rather by admins?

eahub/profiles/management/commands/profilesfromlegacyhub.py
```diff
@@ -7,6 +7,8 @@
 from django.utils import html, timezone
```
sebgrebe: I wonder whether we can get rid of this code. The migration from legacy hub has happened, not sure it's worth our effort to maintain this code

eahub/profiles/tests/test_tags_api.py
```diff
@@ -0,0 +1,87 @@
...
+        response = self.client.get(self._url_detail(profile.pk))
+        self.assertEqual(response.status_code, 200)
+        tags = response.data[tags_field_name]
+        self.assertIn(tag1.pk, response.data[f"{tags_field_name}_pks"])
+        self.assertIn(tag2.pk, response.data[f"{tags_field_name}_pks"])
+        for tag in tags:
+            if tag["name"] == tag1.name:
+                self.assertEqual(tag["types"][0]["type"], type_enum.value)
+
+    def _test_deletion(self, type_enum: ProfileTagTypeEnum):
```
sebgrebe, line 62: Is this test dependent on test_addition having run first? And is this test order guaranteed?

sebgrebe (changes requested):
Initial review with clarificatory questions. Once we have speced this out I can do another review (also once  the bs5 PR is in, bit tricky to review this as the changes are in here too)

eahub/config/settings.py
viktor-yunenko: "FeaturePolicyMiddleware" has been depracated 1-2 years ago, it's now called "permissions"

eahub/profiles/api/views.py
viktor-yunenko: here I would need to add better error handling for the sake of UX, for now the user would just be unable to add the tag and wouldn't see any error message, which would be frustrating

eahub/profiles/management/commands/profilesfromlegacyhub.py
viktor-yunenko: I agree here, I've been thinking about scheduling a 1-2 week-long cleanup once we're done with the pending tasks and have a bit of time. Ie to follow the best practice of refactoring sprints. 

I would say the only reason I haven't been advocating for one is that we're still in the project feasibility assessment stage.

eahub/profiles/tests/test_tags_api.py
viktor-yunenko: not really - django runs each test in a db transaction, hence here django regenerates the tags for every test function

eahub/config/settings.py
viktor-yunenko: they're needed in few templates. In short it's perfectly ok to add to this list everything, except the secrets as `RECAPTCHA_SECRET_KEY` which must never be rendered through HTML

eahub/base/static/components/mailchimp/mailchimp.js
viktor-yunenko: I have no idea really, we've quite a bit of dead code around, afaik this is one of those pieces.

eahub/base/static/components/profile-edit-image.js
viktor-yunenko: I agree, I would be happy to look into it during the profile editing form composition

eahub/profiles/api/serializers.py
viktor-yunenko: the idea of mine was to unblock users - eg if somebody thinking that `couchsurfing` tag is a must, now instead of squeezing it into something nonsensual as "expertise area" they can add a tag as `Helping with couchsurfing" or "Looking for couchsurfing". After that we can notice those tags, analyze the demand and consider creating a better organized field or a set of fields.

I would say if we were god-like human beings who knew everything about our users' desires we wouldn't need a field as `tags_generic` :)

sebgrebe:
[tasks moved to the PR description]

eahub/base/static/components/profile-edit-image.js
viktor-yunenko: all logic form this file was replaced with django-select2 and native bootstrap5 styles

eahub/profiles/api/serializers.py
viktor-yunenko: I'm planning to import the values from EAG "Affinity" field, eg "People managers", "Earning to give", "Parenting", etc

vagarwalla:
Cause area tags

Can we allow people to suggest them but it isn't visible unless an admin approves it?
Otherwise, do not allow for now. Maybe other causes can go in the general tag box later one.

Reasoning: 
- If we tried to integrate that data with CEA's data in the future for any reason, it would probably be messy

viktor-yunenko:
As I mentioned before I'm neutral about preventing the creation of cause areas tag.

I wouldn't like to base our decisions on the fields sharing though, mostly after working with the data that we've got from EAG.

Getting the data is amazing btw, thank you Vaidehi a lot for organizing it!

We've been maintaining the fields to correspond to what we thought CEA was collecting, but in the latest export after I exclude the generic fields (email, name, country/city), we're left with only one kind-of-compatible-field - Local Groups.

One can say perhaps two compatible fields, if we're willing to assume that `Area of expertise/study` is directly importable, because it doesn't fit naturally into only `Expertise areas` or `Career interest`.

In short I don't expect EA organizations to have a stable idea of what fields they want to have, most likely it's going to change by the time of our next import.

viktor-yunenko:
> If we tried to integrate that data with CEA's data in the future for any reason, it would probably be messy

Hence yes, right now it's as messy as it can get :)

Not a significant issue for me though, as a data import isn't a big deal. Fixing UX problems caused by pseudo-compatibility could be a major issue for us though.

viktor-yunenko:
Hi @sebgrebe, I didn't test the logging (which probably doesn't work) and will be doing eg multi-browser testing later. But you can already start reviewing the code. Would you like to do it over 1-2 calls? I can be around eg on Saturday.

eahub/localgroups/forms.py
```diff
@@ -21,12 +22,11 @@ def __init__(self, *, user, local_group, **kwargs):
...
-            queryset = queryset.filter(
-                models.Q(profile__is_public=True)
-                | models.Q(already_selected=True)
-                | models.Q(pk=user.pk)
-            )
+        queryset = queryset.filter(
+            models.Q(profile__is_public=True, profile__is_approved=True)
+            | models.Q(already_selected=True)
+            | models.Q(pk=user.pk)
+        )
```
viktor-yunenko, line 29: I wasn't sure why "superuser" here is important, hence I removed it for simplicity sake, otherwise it's hard to test since I'm seeing bots in the dropdown as an admin. I recommend to do anything admin related in the admin panel.

eahub/base/static/components/profile/edit/profile-edit.ts
```diff
@@ -0,0 +1,152 @@
...
+    private tagsCreateUrl = `/profile/api/profiles/tags/create/`;
+    private http = new HttpService();
+
+    async mounted() {
+        const response = await this.http.get(this.tagsUrl);
+        this.tagsSelected = response.data[`tags_${this.typeName}`];
+
+        // todo fix vue nextTick bug, ie with what/vert snippet I wrote for Vlad
+        await this.sleep(300);
+        this.typesRef.refine(this.typeName);
```
viktor-yunenko: I added it to the todo

viktor-yunenko:
I also added the logs for tags and local groups.

docker-compose.yml
```diff
@@ -1,4 +1,4 @@
-version: "3.4"
+version: "2"
```
sebgrebe, line 1: why?

eahub/templates/eahub/edit_profile.html
```diff
@@ -1,81 +1,215 @@
...
+            <div class="col">
+              {% include 'includes/field.html' with form=form rows=2 field=form.looking_for placeholder="Chat about wild animal suffering" %}
+            </div>
+          </div>
+
+
+          <h2>Community</h2>
+          {% include 'includes/field.html' with form=form field=form.available_as_speaker %}
+
+          {% include 'includes/field.html' with form=form rows=1 field=form.topics_i_speak_about %}
```
sebgrebe: Tag "Topics I speak about" missing

eahub/profiles/migrations/0022_add_field_last_name.py
```diff
@@ -0,0 +1,95 @@
...
+            model_name="profile",
+            name="facebook_url",
+            field=models.URLField(blank=True, max_length=400, verbose_name="Facebook"),
+        ),
+        migrations.AlterField(
+            model_name="profile",
+            name="is_public",
+            field=models.BooleanField(
+                default=True,
+                help_text="Unchecking this will completely conceal your profile",
```
sebgrebe, line 56: I'd add "Public profiles are visible on the web."

eahub/templates/eahub/edit_profile.html
```diff
@@ -1,81 +1,215 @@
...
+              {% include 'includes/field.html' with form=form field=form.city_or_town %}
+            </div>
+          </div>
+          {% include 'includes/field.html' with form=form field=form.image %}
+
+          {% include 'includes/field.html' with form=form rows=7 field=form.summary placeholder="In West Philadelphia born and raised. On the playground is where I spent most of my days." %}
+
+          <div class="mb-5"
+            data-type-name="generic"
+            data-type-label="Tags"
```
sebgrebe: Maybe rename to "Other tags"? @vagarwalla, any native speaker suggestions? We currently use "Tags" for all tags we haven't assigned to categories like "Cause Area", i.e., it's a dustbin category. Other idea would be "Miscellaneous"

eahub/localgroups/tests/test_localgroups_models.py
```diff
@@ -51,34 +51,3 @@ def test_organisers_names_handles_users_without_profiles(self):
         organisers_names = local_group.organisers_names()
 
         self.assertEqual("User profile missing", organisers_names)
-
-    def test_get_exportable_field_names(self):
```
sebgrebe: was this removed accidentally?

eahub/templates/eahub/edit_profile.html
```diff
@@ -1,81 +1,215 @@
...
+          </div>
+
+          {% if profile.cause_areas_other %}
+            {% include 'includes/field.html' with form=form rows=1 field=form.cause_areas_other %}
+          {% endif %}
+        
+
+          <h2>Settings</h2>
+          {% include 'includes/field.html' with form=form field=form.is_public %}
+          {% include 'includes/field.html' with form=form field=form.allow_messaging %}
```
sebgrebe, line 203: add note that email won't be visible to other users

sebgrebe (changes requested):

docker-compose.yml
sebgrebe: Answered: Because we're actually using version 2

eahub/templates/account/signup.html
```diff
@@ -33,11 +33,21 @@ <h1>Sign Up</h1>
...
+
+            <div class="form-floating">
+                <input type="text"
+                    class="form-control form-control-lg"
+                    name="{{ form.last_name.name }}"
+                    placeholder="Full Name"
+                    id="{{ form.last_name.id }}"
+                    required>
+                <label for="{{ form.last_name.id }}">Last Name</label>
             </div>
```
sebgrebe, line 51: Needs to be cleaned up: 
image

Placeholders "FullName" are confusing

sebgrebe: Same for login (I did a hard refresh)
image

eahub/templates/eahub/edit_profile.html
```diff
@@ -1,81 +1,215 @@
...
+          >
+            <div class="vue-profile-tag-input"></div>
+          </div>
+
+          {% if profile.expertise_areas_other %}
+            {% include 'includes/field.html' with form=form rows=1 field=form.expertise_areas_other %}
+          {% endif %}
+
+          <div class="mb-5"
+            data-type-name="cause_area"
```
sebgrebe, line 170: On QA session with @vagarwalla we noticed that there are some non-cause areas listed here, like "Software Engineering" and "Giving What We Can"

eahub/localgroups/tests/test_localgroups_models.py
viktor-yunenko: restored 👌

eahub/templates/account/signup.html
viktor-yunenko: a caching issue

sebgrebe: Turned out I had to rerun ```npm ci``` because of new dependencies, so it served the old build

eahub/templates/eahub/edit_profile.html
```diff
@@ -1,81 +1,224 @@
...
+                </a>
+              </div>
+            </div>
+          </div>
+          <div class="row row-cols-2">
+            <div class="col">
+              {% include 'includes/field.html' with form=form field=form.country %}
+            </div>
+            <div class="col">
+              {% include 'includes/field.html' with form=form field=form.city_or_town %}
```
sebgrebe: I think it's standard to have city/town before country

eahub/templates/eahub/edit_profile.html
```diff
@@ -1,81 +1,224 @@
...
+          {% include 'includes/field.html' with form=form field=form.available_as_speaker %}
+
+          {% if profile.topics_i_speak_about %}
+            {% include 'includes/field.html' with form=form rows=1 field=form.topics_i_speak_about %}
+          {% endif %}
+          <div class="mb-5"
+            data-type-name="speech_topic"
+            data-type-label="Topics I speak about"
+          >
+            <div class="vue-profile-tag-input"></div>
```
sebgrebe, line 109: image
This sounds to me a bit like "Never create a tag". Suggestion: "Try to find an existing tag before creating a new one"

eahub/templates/eahub/edit_profile.html
```diff
@@ -1,81 +1,224 @@
...
+          </div>
```
sebgrebe, line 110: image

EAHub is missing :) @vagarwalla could you do a systematic check on your QA whether we still have all the items on each list (cause area, career interest, skills, organisational affiliation) we used to have?

eahub/templates/eahub/edit_profile.html
```diff
@@ -1,81 +1,224 @@
...
+    {% endif %}
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/select2/4.0.12/js/select2.min.js" defer></script>
+    <script src="{% static 'django_select2/django_select2.js' %}" defer></script>
+    {{ form.media.css }}
+  </div>
 
 {% endblock %}
 
-{% block submit%}Update{% endblock%}
+{% block footer %}
```
sebgrebe, line 238: Footer missing

image

eahub/templates/eahub/edit_profile.html
```diff
@@ -1,81 +1,224 @@
...
+          </div>
+          
+          <div class="mb-5"
+            data-type-name="pledge"
+            data-type-label="Pledges"
+          >
+            <div class="vue-profile-tag-input"></div>
+          </div>
+
+          <div class="mb-5"
```
sebgrebe, line 130: I only get this locally, not on the deployed server: On selecting an event, this occurs:
image

Any idea how this could be fixed? Not urgent prior to release, but ideally we have the same functionality locally

eahub/templates/eahub/profile.html
```diff
@@ -124,113 +145,128 @@ <h1 class="col col-auto">{{ profile.name }}</h1>
         <div class="prof__summary">
           {{ profile.summary|default:''|urlize|linebreaks }}
 
-          <div class="row summary__sections">
-            {% if profile.offering %}
```
sebgrebe, line 128: offering and looking for not on same height

eahub/templates/eahub/profile.html
```diff
@@ -39,81 +39,102 @@
...
-                ></i> Change Email Address</span>
-              </a>
-              <a class="btn btn-outline-primary btn-sm"
-                href="{% url 'account_change_password' %}"
-              >
-                <span><i class="fa fa-pencil-alt"></i> Change Password</span>
-              </a>
-              <a class="btn btn-outline-primary btn-sm"
-                href="{% url 'edit_profile' %}"
+            <div class="prof__btns col col-24 d-none d-lg-flex">
```
sebgrebe, line 135: (optional) I think Edit button and social media buttons should be at same height, looks a bit wonky.
If you don't want to do this now, you could add this to some list for to be done later

image

eahub/profiles/models.py
```diff
@@ -482,220 +296,58 @@ def geocode(self):
...
-        self, enum_values_list: List[Union[enum.Enum, str, int]], enum_cls: enum.Enum
-    ) -> List[str]:
-        enum_labels: List[str] = []
-        for enum_value_raw in enum_values_list:
-            enum_value = int(enum_value_raw)
-            enum_labels.append(enum_cls.values[enum_value].label)
-        return enum_labels
-
-    @staticmethod
-    def get_exportable_field_names():
```
sebgrebe, line 693: Export on profiles misses a number of fields, among others:
- Organisational affiliation
- speech topic
- open to volunteer
- social media pages

The custom code you removed was relying on this method here to get all those fields, and the 3rd party library we're using now seems to be missing them.

sebgrebe (changes requested):

eahub/profiles/models.py
viktor-yunenko: I know, we don't need to handle those fields in a custom manner anymore, as they're handled automatically when added to the `fields = []` attribute of the resource class.

I added few more fields that you mentioned, if you want more you can add them as well.

Does that sound sufficient?

eahub/templates/eahub/edit_profile.html
viktor-yunenko: Did you clone the database from the tags branch before? This list comes from algolia index from the tags db, it won't work without the identical SQL rows.

eahub/templates/eahub/profile.html
viktor-yunenko: fixed

eahub/templates/eahub/edit_profile.html
viktor-yunenko: I think it was standard with paper latters to make it easier for the postman to shuffle through those? which doesn't sound as an appealing standard base to me

I can change it if you're sure

eahub/templates/eahub/profile.html
viktor-yunenko: yes, it's a firefox bug

FYI I develop only on chrome, hence until multi-browser testing is done I don't look at other browser

eahub/templates/eahub/edit_profile.html
viktor-yunenko: moved up to the todo list in the ticket

eahub/profiles/models.py
sebgrebe: Fine for now, when we want to export data to other parties we probably want all fields, but we can change that then

eahub/base/static/components/profile/edit/profile-tag-input.ts
```diff
@@ -0,0 +1,158 @@
...
+            tag => tag.pk !== Number(pkToDrop)
+        );
+        try {
+            let data = {};
+            data[`tags_${this.typeName}_pks`] = tagsSelectedNew.map(tag => tag.pk);
+            await this.http.patch(this.tagsUrl, data);
+            this.tagsSelected = tagsSelectedNew;
+        } catch (e) {
+            tagToDrop.isLoading = false;
+            alert('An error occurred');
```
sebgrebe: optional: add error message to alert

eahub/localgroups/forms.py
sebgrebe: Nice, this seems to have fixed this issue! #1190

eahub/templates/eahub/edit_profile.html
sebgrebe: Had to look this up, it seems that what's conventional differs on world regions: https://ux.shopify.com/designing-address-forms-for-everyone-everywhere-f481f6baf513

As I'm used to the western system, I found the order you used counterintuitive, but other users will differ.

I would suggest though that we should keep it aligned with the groups editing form. So either we need to swap it around there, too (country, city) or we put it back to the original order here (city, country). Both works for me.

eahub/templates/eahub/talentsearch.html
```diff
@@ -41,10 +41,10 @@
           {% thumbnail profile.image "200x200" crop="center" as im %}
```
sebgrebe: We should remove this page altogether! We forgot to remove it when we introduced the new profiles search:
- https://eahub.org/volunteers/  
- https://eahub.org/speakers/  
- https://eahub.org/candidates   

Can be done in a separate MR (but let's create a ticket to not forget)

sebgrebe (changes requested):

eahub/templates/eahub/talentsearch.html
viktor-yunenko: always happy to drop dead code - done 👍

vagarwalla:
1. Cause Area Mapping done here: https://docs.google.com/spreadsheets/d/1Y_GKftJeWh5LBYflnPJwrFrLN6BgIiTGzM-CA-fGK8I/edit#gid=189592928
2. Tag Field Descriptions: in the same doc in a different tab.

eahub/templates/eahub/edit_profile.html
viktor-yunenko: for ux reasons as discussed

viktor-yunenko:
I moved the description discussion to a google doc to simplify commenting - https://docs.google.com/document/d/1KnoedOwlJGqdzHSkCtqhJpBJX-WKUcUPUWUhU1leUaE/edit

eahub/templates/eahub/edit_profile.html
viktor-yunenko: done

eahub/profiles/migrations/0022_add_field_last_name.py
viktor-yunenko: done

viktor-yunenko:
I finished processing the changes. @sebgrebe, please feel free to check. I'll proceed with the deployment at 11-12 by GMT if nobody minds.

viktor-yunenko:
We might want to discuss labels in detail based on my google doc comments, but I'm assuming those aren't blockers?

eahub/templates/eahub/edit_profile.html
sebgrebe: this still seems to be missing, @victor-yunenko? And did we do a check that all tags have been successully moved? Wondering because if EAHub is missing, other things might be missing, too. Or does the tags server not have the complete db?

viktor-yunenko: afaik it's called Effective Altruism Hub

but we of course should add synonyms as "eahub"

viktor-yunenko: and yes I've been testing the enums migration few times, as I needed to reset the database and rerun the migrations every week or something

so far the only issue we had is that I thought that "cause areas" are the same as "expertise areas", but it was just a silly assumption rather than a real bug

PR #1180, opened 2021-02-24
1018 add feedback box

LisaJD:


sebgrebe (changes requested):
Please put all changes into one MR, as discussed

Issue #1186, opened 2021-02-28
Move Privacy Policy page to bootstrap 5


sebgrebe:
### Current state
* On some pages we use bootstrap 3, on some bootstrap 5. We want all on 5
* Privacy policy oage is still on bootstrap 3

### Desired state
* Provacy policy page uses bootstrap 5: https://eahub.org/privacy-policy/
* Example of page that has already bootstrap 5: https://eahub-stage.us.aldryn.io/about/

### Acceptance criteria
* [ ] Privacy policy page uses bootstrap 5

viktor-yunenko:
That was done

PR #1208, opened 2021-03-11
Upgrade privacy policy to bootstrap5

walambert:
#1186 
(https://github.com/rtcharity/eahub.org/issues/1186)

Upgraded privacy policy page to bootstrap5 by making privacy_policy.html template extend base_bs5.html instead of base.html.

Added navbar block as an include or else it wouldn't display properly after page upgrade to bs5; added line break before h1 so that spacing between navbar and content was appropriate.


eahub/templates/eahub/privacy_policy.html
```diff
@@ -1,9 +1,15 @@
-{% extends 'base.html' %}
+{% extends 'base_bs5.html' %}
 
 {% block title %}Privacy Policy{% endblock %}
 
+<!--navbar block required include after bs3 to bs5 upgrade for page to format correctly-->
```
sebgrebe: The navbar and the footer still look like on bs3. It should look like on the About page: https://eahub.org/about/

sebgrebe (changes requested):

sebgrebe (approved):

PR #1225, opened 2021-04-02
Add feedback box

### LINKED ISSUE #1018: Add a feedback box for the profiles search
* An example of the functionality can be found here - https://optimus.foundation/en/projects/13/ (The email notification isn't necessary for now, as well as the up/down buttons.)
* Only to be added to eahub.org/profiles
* Data should be saved in django model
  * Store page url, too. Atm, this will only be /profiles, but we'd like it to be extandable for the future
* Data should be exportable from admin panel
* I would recommend a bootstrap4 modal window with email (optional) and message (multiline, required) field.
* On mobile perhaps it can be reduced in size compared to the optimus.foundation version?
* Feedback button should fit with eahub's color scheme

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1199658157929710/1199881547599376) by [Unito](https://www.unito.io/learn-more)

sebgrebe:
closes #1018


eahub/base/static/components/feedback.js
```diff
@@ -0,0 +1,26 @@
+async function ajaxsubmit(){
+  document.getElementById('page_url').value = window.location.href;
```
viktor-yunenko: I think it's better to either do that on the backend or replace with `window.location.pathname`, because `href` adds any user specific parameters that will look odd in admin, eg

window.location.href
"http://127.0.0.1:8000/profiles/?toggle%5Bavailable_to_volunteer%5D=true&refinementList%5Bcountry%5D%5B0%5D=USA"

viktor-yunenko: while window.location.pathname is `/profiles/`

eahub/base/static/components/feedback.js
```diff
@@ -0,0 +1,26 @@
...
+    body: new URLSearchParams(new FormData(form)),
+    headers: {
+      'Content-Type': 'application/x-www-form-urlencoded'
+    }
+  }).then(response => {
+    if(response.status == "200") {
+      document.getElementById("successmsg").style.display = "block";
+      document.getElementById("form").style.display = "none"; 
+    }
+    if(response.status != "200") {
```
viktor-yunenko: could we also replace both of those with `===`?

eahub/base/static/global/styles/components/_feedback.scss
```diff
@@ -0,0 +1,35 @@
...
+  transform: rotate(-90deg);
+  transform-origin: top left;
+  color: #fff;
+  text-align: center;
+  right: -65px;
+} 
+
+a {
+  color: inherit;
+}
```
viktor-yunenko: those would be global styles? I'm not sure how all browsers would process that

could we encapsulate it in a component class?

`<div class=feedback-componenent>`

```
.feedback-componenent {
    // other styles
}
```

viktor-yunenko (approved):

viktor-yunenko:
Thanks everyone! I just merged it in 👍

PR #1243, opened 2021-04-12
EAG import fine tuning and password reset flow cleanup

viktor-yunenko:
Hi @sebgrebe, I prepared few updates, especially with the password reset flow.

We need to deploy this before the imports.

You can check them here:
- https://eahub-eagimport-b73ae45.us.aldryn.io/accounts/password/reset/
- https://eahub-eagimport-b73ae45.us.aldryn.io/accounts/password/reset/done/
- https://eahub-eagimport-b73ae45.us.aldryn.io/accounts/password/reset/key/mas-5pp-4a675f08a867cb2a5659/

PR #1244, opened 2021-04-18
1120 internal profiles

sebgrebe:
Adds public/internal/private profile visibility:  
  - public profiles on search and profile page are visible to anyone    
  - internal profiles on search and profile page are visible only to logged in, approved users

Stage server: https://eahub-visibility-ae3e261.us.aldryn.io

### Changes  
- Adds new enum field on profile model for visibility  
- Adds migration logic that maps old field ```is_public``` to new ```visibility``` field  
- Adds second index for internal profiles (following [this pattern](https://www.algolia.com/doc/framework-integration/django/options/?client=python#multiple-indices-per-model))
- On the map, only public profiles are shown with profile information (internal ones are counted as private) [We can change this in a subsequent PR such that logged in users can see internal users on the map]
- On the pages of local groups, only public users are listed as organisers [We can change this in a subsequent PR such that logged in users can see internal users as organisers]
- Replaces "is_public" field on admin panels with "visibility"


eahub/profiles/index.py
```diff
@@ -51,6 +50,49 @@ class ProfileIndex(AlgoliaIndex):
...
+        should_index = "is_searchable_internal"
+        fields = ProfilePublicIndex.fields
+
+    @register(Profile)
+    class ProfileMetaIndex(AlgoliaIndex):
+        def __init__(self, model, client, settings):
+            self.indices = [
+                ProfilePublicIndex(model, client, settings),
+                ProfileInternalIndex(model, client, settings),
+            ]
```
viktor-yunenko, line 65: @sebgrebe , could you clarify the advantage over using two different classes that would index based on `should_index` method value?

I'm asking because we've spent a lot of time struggling with `algoliasearch_django`, and found out that extending it is a long-term maintenance issue.

eahub/profiles/models.py
```diff
@@ -114,7 +120,7 @@ def visible_to_user(self, user):
         if user.is_superuser:
             return self.all()
         return self.filter(
-            models.Q(is_public=True, is_approved=True) | models.Q(user_id=user.pk)
+            models.Q(is_publicly_visible=True, is_approved=True) | models.Q(user_id=user.pk)
```
viktor-yunenko: should we replace all occurrences of the old variable with `exclude(visibility=Visibilty.PRIVATE)` or something? because I'm assuming this flag could become outdated now?

eahub/profiles/tests/test_views.py
```diff
@@ -0,0 +1,41 @@
+from django.test import override_settings
+
+from eahub.tests.cases import EAHubTestCase
+from eahub.profiles.models import VisibilityEnum
+
+
+@override_settings(STATICFILES_STORAGE='django.contrib.staticfiles.storage.StaticFilesStorage')
```
viktor-yunenko: [optional] a good one! probably we can move this decorator to `EAHubTestCase`?

eahub/profiles/views.py
```diff
@@ -162,4 +163,9 @@ def delete_profile(request: HttpRequest) -> HttpResponse:
 
 
 def profiles(request) -> HttpResponse:
-    return render(request, "eahub/profiles.html", {"feedback_form": FeedbackForm()})
+    approved_user = False
+    if request.user.is_authenticated:
+        profiles = Profile.objects.all().filter(user=request.user)
+        if len(profiles) == 1 and profiles[0].is_approved:
+            approved_user = True
+    return render(request, "eahub/profiles.html", {"feedback_form": FeedbackForm(), "approved_user": approved_user })
```
viktor-yunenko: could this be replaced with `{{ request.user.profile.is_approved }}` in html? looks like it could save few lines of python then, since django template fallbacks to render an empty string?

eahub/profiles/tests/test_views.py
```diff
@@ -0,0 +1,41 @@
...
+        profile.visibility = VisibilityEnum.PRIVATE
+        profile.save()
+        response = self.client.get('/profile/first-last/')
+        self.assertEqual(response.status_code, 404)
+
+    def test_internal_profile_hidden(self):
+        profile = self.gen.profile(first_name="first", last_name="last")
+        profile.visibility = VisibilityEnum.INTERNAL
+        profile.save()
+        response = self.client.get('/profile/first-last/')
```
viktor-yunenko: FYI optional, but could also be reversed, eg as `reverse("profiles_app:profile", {"slug"=profile.slug})`

viktor-yunenko: I also sometimes just create a method on the model as `def get_url(self) -> str`

eahub/profiles/tests/test_views.py
```diff
@@ -0,0 +1,41 @@
...
+        profile_visitor.is_approved = True
+        self.client.force_login(profile_visitor.user)
+
+        response = self.client.get('/profile/first-last/')
+        self.assertEqual(response.status_code, 200)
+
+    def test_public_profile_visible(self):
+        profile = self.gen.profile(first_name="first", last_name="last")
+        profile.visibility = VisibilityEnum.PUBLIC
+        profile.save()
```
viktor-yunenko, line 50: [optional] I wonder whether visibility could be put in `gen.profile()` call to save few lines?

viktor-yunenko (commented):
thank you for the quick help with this one! just left few questions

eahub/profiles/index.py
viktor-yunenko: I saw that it's part of the docs, and of course following the official advice would be ideal. `algoliasearch_django` is a special case though, I've spent few months studying its codebase, and now I'm very hesitant to apply something that is listed as a best practice there 😅

eahub/profiles/models.py
sebgrebe: ```is_publicly_visible``` is set to ```visibility=Visibility.PRIVATE or visibility=Visibility.INTERNAL``` so means the same as the exclude code. 

But I don't care too much

eahub/profiles/tests/test_views.py
sebgrebe: Weirdly, ProfileTestCase does not seem to inherit the decorator from EAHubTestCase. I got a staticfiles error when I removed it and added it to EAHubTestCase instead

eahub/profiles/views.py
sebgrebe: done

eahub/profiles/tests/test_views.py
sebgrebe: done

eahub/profiles/tests/test_views.py
sebgrebe: done

eahub/profiles/index.py
sebgrebe: Do you mean having two different Profile classes, one for internal, one for public? Wouldn't that great quite a bit of overhead, e.g., you'd need 2 tables? And wouldn't having more than 1 profile model create maintainability issues?

I've just pushed some code with a naive first implementation of what I think you have in mind, but it didn't work because the "ProfilesInternal" table is empty.

viktor-yunenko: ok, the error makes sense

eahub/profiles/models.py
viktor-yunenko: I wouldn't care about it either, but loading objects in RAM instead of a single SQL query is usually hard for django and doesn't go unnoticed. I'll create a ticket then.

sebgrebe:
Closing as [this PR](https://github.com/rtcharity/eahub.org/pull/1243) was rebased on the present one

PR #1275, opened 2021-06-01
Added step-by-step Contribution Tutorial to CONTRIBUTING.md

walambert:
I wrote a step-by-step contribution tutorial and added it to our [contribution guide ](https://github.com/rtcharity/eahub.org/blob/21b6111b2626e4739c249d0881d16fbc818094cb/CONTRIBUTING.md) to aid with onboarding new volunteers and contributors, especially if they don't have much experience using git or GitHub.

Cheers,
Willa


CONTRIBUTING.md
```diff
@@ -67,3 +67,71 @@ squashed one:
 1. Do not end the subject line with a period
 1. Use the imperative mood in the subject line
 1. Use the body to explain what and why vs. how  
+
+## Contribution Tutorial
```
sebgrebe, line 71: We're wondering whether it would maybe better add this tutorial to some google doc we could share or we link to this document on your branch. Our concern is that these instructions could go out of date if git changes things. As it is a standard action done on git, developers should generally either know it already or be able to google it. Hence, there is probably no need to add it to our own documenation.

CONTRIBUTING.md
```diff
@@ -38,7 +38,7 @@ join the [contributions channel](https://discord.gg/CQueVjk3fc) on EA Hub's disc
 that you'd like to contribute.  
 
 ### Pull Requests  
-To make a pull request, please fork the repo, create a branch on your fork and then make a PR into the main repo.
+To make a pull request, please fork the repo, create a branch on your fork and then make a PR into the main repo. For a detailed tutorial, please see our Contribution Tutorial at the end of this page.
```
sebgrebe, line 41: Thanks for fixing all the language issues! :)

sebgrebe (changes requested):

CONTRIBUTING.md
walambert: Would it be better to describe what actions are being done and link to GitHub's or Git-SCM's own documentation about how to perform those actions? Some contributors are relatively new to using Git and having some space for Git specific instructions would help facilitate their onboarding (plus future contributors) to making contributions, I believe.

Also, the tutorial should include our project / context specific instructions such as the style guide already part of this doc, CONTRIBUTING.md as well as "you must do this before making a pull request" instructions from the [SETUP.md ](https://github.com/rtcharity/eahub.org/blob/default/SETUP.md) Formatting section, yes?

Should we duplicate or move the Formatting section of SETUP.md to the contribution tutorial so that when making a pull request, one doesn't have to check multiple pages?

I worry that making a google doc spreads the documentation further afield and makes a contributor check even more things to get started, and would have the same issues of potentially going out of date if git changes things too.

CONTRIBUTING.md
walambert: Cheers :) Let me know if there are other pages I should check for such things.

sebgrebe:
Closing since the best place for this seems to be outside of version control (see comment [here](https://github.com/rtcharity/eahub.org/pull/1275#discussion_r646135576))

PR #1285, opened 2021-06-18
1283 reindex on approval

sebgrebe:
- Fixes bug that prevented reindexing of profiles on approval of profile  
- Removes profile approval action from user admin panel: it's a profile action, not a user action, so should only be on the profiles panel)  

Tested on stage


viktor-yunenko (approved):
ah, one more of django ORM quirks, thank you for figuring it out!

PR #1292, opened 2021-08-08
1159 - have messages only in base template

LisaJD:
(cherry picked from commit 6cf2cd62ddc4e294696abdeec0e65bdb2ee1069f)


eahub/templates/base_bs5.html
```diff
@@ -21,3 +21,18 @@
     <link  rel="stylesheet" href="{% static 'dist/global_bs5.css' %}">
   {% endif %}
 {% endblock static %}
+
+{% block messages %}
+
+{% if messages %}
+<div class="container mt-2 w-75">
+  {% for message in messages %}
+  <div class="alert {% for tag in message.tags %} alert-{{ tag }} {% endfor %} alert-dismissible fade show" role="alert">
```
sebgrebe, line 30: * Testing this, I've noticed that the loop is incorrect here. ```message.tags``` is not a set of values, but just one (e.g., "success"), so the loop ends up looping over letters. The result of this is that the css is messed up and you actually have no styling on the message.

You can just do ```alert-{{message.tags}}``` here.

* this is missing the hub-{{message.tags}} (the bootstrap default color doesn't match the hub colour pallette)

eahub/templates/base_bs3.html
```diff
@@ -17,3 +17,20 @@
     <link  rel="stylesheet" href="{% static 'dist/global_bs3.css' %}">
   {% endif %}
 {% endblock static %}
+
+{% block messages %}
+
+{% if messages %}
+<div class="container" style="padding-top: 20px; max-width: 70%">
```
sebgrebe, line 24: Remove inline styling as in bs5 template

eahub/templates/base_bs5.html
```diff
@@ -21,3 +21,18 @@
...
+    {{ message }}
```
sebgrebe, line 31: There is a problem with including messages into the base template on the ```accounts/confirm-email/``` page appearing after signing up since we have hardcoded the message in that case ("Confirmation email sent to <email-address>") into the html page for ```accounts/confirm-email/```.  

Please remove that hard coded message so that only the dynamic message shows.

Another problem is that for this message the bootstrap class "alert-info" is shown, but for some reason no styling is applied here. Please do either of:
- Create ```hub-info``` class with a neutral background-color  
- Figure out why ```alert-info``` is not applied and fix it

image

sebgrebe (changes requested):

sebgrebe:
closing this as progress has been stalling

PR #1297, opened 2021-10-16
Add script for importing users from fellowship weekend

sebgrebe:


eahub/base/management/commands/fellowship_import.py
```diff
@@ -0,0 +1,71 @@
...
+                tag_name = header.replace("tags_", "")
+                profile_tag_type_enum = ProfileTagTypeEnum(tag_name)
+                profile_tag_types = ProfileTagType.objects.filter(type=profile_tag_type_enum)
+                profile_tag_type = profile_tag_types[0]
+                if not profile_tag_types:
+                  continue
+                if not isinstance(value, str):
+                  continue
+                for single_value in [x.strip() for x in value.split(",")]:
+                  profile_tags = ProfileTag.objects.filter(name=single_value)
```
viktor-yunenko, line 42: probably might be case insensitive?

viktor-yunenko: ie `name__iexact=single_value`

sebgrebe: Isn't that a good thing? I.e., this way we don't duplicate tags just because of differences in lower/upper case.

sebgrebe: Oh, that's what you mean. Will change (won't make a difference though as the I don't think there are differences in case in the underlying data)

viktor-yunenko: I agree that there shouldn't be a difference in the fellowship data, but on that matter there's quite a variety in our own dataset 😅

PR #1315, opened 2022-08-22
upgrade to Django 3.2

sebgrebe:
- Updates to Django 3.2.15 and updates python packages that needed updating due to the Django update  
  - As part of this, removes dependency on aldryn django and uses Django directly  
- Replaces aldryn storage with S3Boto3Storage (as aldryn is no longer used)    
  - As part of this, introduces parse_storage_url in config/utils, as we previously used this method from aldryn django, which is no longer used. Our own method is a simplified version of it.   
- Updates docker container to node14 and python 3.9 since at least one of the updated python packages required python 3.9 and there was no docker image with python 3.9 and any node lower than 14  
- Updates model code using enum package as higher version used now required that


eahub/config/settings.py
```diff
@@ -210,17 +211,14 @@ class DjangoEnv(Enum):
...
-    AWS_MEDIA_SECRET_ACCESS_KEY = media_config["AWS_MEDIA_SECRET_ACCESS_KEY"]
-    AWS_MEDIA_STORAGE_BUCKET_NAME = media_config["AWS_MEDIA_STORAGE_BUCKET_NAME"]
-    AWS_MEDIA_STORAGE_HOST = media_config["AWS_MEDIA_STORAGE_HOST"]
-    AWS_MEDIA_BUCKET_PREFIX = media_config["AWS_MEDIA_BUCKET_PREFIX"]
-    AWS_MEDIA_DOMAIN = media_config["AWS_MEDIA_DOMAIN"]
+    DEFAULT_FILE_STORAGE = "storages.backends.s3boto3.S3Boto3Storage"
+    AWS_ACCESS_KEY_ID = media_config["AWS_MEDIA_ACCESS_KEY_ID"]
+    AWS_SECRET_ACCESS_KEY = media_config["AWS_MEDIA_SECRET_ACCESS_KEY"]
+    AWS_STORAGE_BUCKET_NAME = media_config["AWS_MEDIA_STORAGE_BUCKET_NAME"]
+    AWS_S3_CUSTOM_DOMAIN = media_config["AWS_MEDIA_CUSTOM_DOMAIN"]
```
viktor-yunenko, line 219: do you need the custom domain btw? not like it would add issues, just curious :)

eahub/config/settings.py
```diff
@@ -210,17 +211,14 @@ class DjangoEnv(Enum):
...
+    AWS_QUERYSTRING_AUTH = True
```
viktor-yunenko: I think you don't need those one when the bucket is set to public? they also present some issues afaik, since those auth keys can expire?

eahub/config/utils.py
```diff
@@ -0,0 +1,34 @@
...
+                      .format(signature_ver))
+
+    config.update({
+      'AWS_MEDIA_ACCESS_KEY_ID': parse.unquote(url.username or ''),
+      'AWS_MEDIA_SECRET_ACCESS_KEY': parse.unquote(url.password or ''),
+      'AWS_MEDIA_STORAGE_BUCKET_NAME': bucket_name,
+      'AWS_MEDIA_CUSTOM_DOMAIN': url.hostname,
+    })
+
+  return config
```
viktor-yunenko, line 34: is that the divio's snippet? If it only splits the url I personally would prefer to add raw env from AWS as `AWS_MEDIA_ACCESS_KEY_ID`. I believe divio team sometimes assigns junior/middle devs to write such snippets, and then they add extra points of failure to the system

eahub/config/settings.py
sebgrebe: it wasn't working without it, yes

eahub/config/utils.py
sebgrebe: As I need the AWS_MEDIA_CUSTOM_DOMAIN (see above), I'd leave it in. I'm also not sure whether divio might rotate the access keys, which would create a failure point if I infer them from the env var set by divio and add them as a further env manually.

eahub/config/settings.py
sebgrebe: thanks, set to false

PR #1339, opened 2023-02-16
Matchmaking and similarity search frontend and HTTP API for backend.

levitation:


eahub/base/static/components/search-profiles/search.html
```diff
@@ -145,7 +145,7 @@ <h5 class="modal-title">Filters</h5>
                   >
                     <div class="row search__hit-summary" v-if="item.summary">
                       <h5>Summary</h5>
-                      <ais-snippet :hit="item" attribute="summary"/>
+                      <ais-snippet :hit="item" attribute="summary"/><!-- TODO roland: replace newlines with <br>? -->
```
viktor-yunenko: FYI I think that isn't possible, algolia doesn't allow spaces in snippets, that's part of their indexing engine tech specs. And there can't be HTML tags in those fields. Could you remove those TODOs then?

viktor-yunenko:
Thank you Roland!

It generally makes sense. But could you remove your local env code to make the tests pass?

Also you don't need to add `  # roland` notes in code, we have inline git history that developers can enable in IDE when they want 👍

eahub/base/static/components/search-profiles/search.html
levitation: Can you please clarify, what do you mean "doesn't allow spaces in snippets"? Would newline characters, tab characters or non-breaking space characters be allowed? Maybe we could use them in place of HTML &lt;br&gt; tags? And then during rendering replace them with &lt;br&gt; tags?

levitation:
Thank you Victor for the quick review! 

Did the fixes you asked: 
- Reverted settings.py; 
- partially reverted webpack.config.js; 
- removed my name from comments on changed lines; 
- also adjusted test_profile_models.py a bit. 

Now the tests pass.

PR #1341, opened 2024-04-01
Upgrade to django42

sebgrebe:
## Changes because of upgrade to Django 4.2:
- Upgrades to Django 4.2
- Upgrades docker-compose to use Postgres 13
- Replaces ```django_pwned_passwords``` with ```pwned-passwords-django``` as the other was no longer working after django upgrade
- Adds ```crispy-bootstrap3``` (needed with v2 of Django-crispy-forms)
- Fixes ```DateRangeFilter``` import
- Fixes loading ```recaptcha``` package
- Uses ```django.urls.re_path``` instead of ```url``` (depracated)
- Fixes ordering of group organisers: by ```last_name```, not ```name```  
- Fixes initialisation of Selenium WebDriver
- Adds ```pytest.ini``` (selenium was failing after upgrade because it was missing)
- Replaces usage of ```django-cmshelpers``` ```send_email``` method (no longer maintained) with Django's default method  
- Added command ```load_seed_data```  
  - Adds an instance of ```SocialApp``` if in local or test environment to ensure that website can load  
  - Without such an instance, the ```provide_login_url``` template tag was throwing an error. Came after upgrading to 0.60  

## Changes to functionality: Groups page
- Removes ```table.js``` as it has outdated JS dependencies, which are a pain to keep updated  
- This breaks the /groups page though. The page is no longer used, however, but immediately redirected to CEA's community page  
- So if in the future /groups should be reinstantiated, either ```tables.js``` needs to be brought back or some other, new code used

##  Other changes  
- Upgrades docker container to node18 because node14 is no longer supported
    - Added ```NODE_OPTIONS=--openssl-legacy-provider``` to ```npm run build```, which is needed with node18
- Upgrades Sentry
- Local groups tests: compares querysets by elements, not count
- Fixes reference of ```django-sluggable``` to specific commit in requirements.in (not .txt)


eahub/localgroups/views.py
```diff
@@ -132,11 +142,19 @@ def form_valid(self, form):
...
+
+        email = EmailMessage(
+            f"{sender_name} sent you a message",
+            message,
+            settings.DEFAULT_FROM_EMAIL,
+            [sorted(recipient.get_messaging_emails(self.request))[0]],
+            [],
+            reply_to=[form.cleaned_data["your_email_address"]]
         )
+        email.attach_alternative(html_msg, "text/html")
```
sebgrebe, line 156: this shouldn't work as it's not EmailMultiAlternatives

eahub/profiles/migrations/0029_auto_20240331_1457.py
```diff
@@ -0,0 +1,48 @@
+# Generated by Django 3.2.15 on 2024-03-31 14:57
```
sebgrebe: django 3.2? Can this be removed?

eahub/tests/e2e/test_signup.py
```diff
@@ -10,6 +10,9 @@ def test_signup(self):
         user_email = "test@eahub.org"
         user_password = "Wa4@;fh>A/~W#6SH"
 
+        print("LIVE SERVER URL")
```
sebgrebe: rm

eahub/profiles/migrations/0029_auto_20240331_1457.py
sebgrebe: Same with localgroups

sebgrebe (commented):
LGTM

